PR #1824, opened 2026-04-30
fix(test-and-mark-stable): consolidate soft-error reports across every smoke-test job via deterministic concat

shubhodeep1:
## Summary

The release-gate's final soft-error summary could read "no issues" even when individual per-phase analyser runs flagged real soft failures. Two root causes:

1. **Coverage gap.** The Phase 8 consolidation lived inside `e2e-smoke-test` and only saw its own 6 phase run IDs. The 11 additional per-phase artifacts produced by sibling smoke-test jobs (`workflow_log_analysis`, `validation_refresh`, `update_workflows`, `memory_maintenance`, `orchestrate_decompose`, `validate_standalone`, `clarify_negative`, `alt_clarify`, `alt_plan`, `alt_implement`, `alt_review_autofix`) never reached the consolidated report or the Telegram notification.
2. **Fidelity loss.** Phase 8 re-ran the OpenRouter analyser across the 6 logs at once with an `<800 words` prompt cap; the model consolidated per-phase findings into "patterns to watch" or dropped them. The Telegram summary was then byte-truncated at 2800 chars, which could chop off real findings depending on section ordering.

This PR replaces the LLM second-pass with a deterministic concat performed in the `notify` job — which already `needs:` every smoke-test sibling and therefore can see every per-phase artifact.

## What changed

- New `scripts/consolidate_soft_error_reports.py`: pure-concat merger. Walks the downloaded artifact directory, sorts phases by canonical pipeline order, concatenates the per-phase markdown verbatim, and assembles a section-aware Telegram summary that drops whole finding blocks (rather than slicing mid-finding) plus a status pill exposing each phase's analyser status code (`ok` / `call_failed` / `api_skipped` / `unreadable`) so analyser failures are visible at a glance.
- `notify` job now downloads `soft-error-report-${{ github.run_id }}-*` (every per-phase artifact in this run), runs the consolidator, uploads the consolidated report under the existing `soft-error-report-<run_id>` artifact name, and reads the section-aware summary file for the Telegram body.
- Removed the redundant Phase 8 LLM re-summarisation, its 5-min wait step, its per-phase env wiring, the byte-truncated `soft_error_summary` job output, and the now-redundant artifact upload from `e2e-smoke-test`. The per-phase analyser invocations remain the source of truth.

## Backward compatibility

The `soft-error-report-${{ github.run_id }}` artifact name is preserved (now contains the consolidated concat instead of the LLM re-summarisation). The Telegram trailer's "see artifact 'soft-error-report-…'" reference still resolves correctly. No external workflows / scripts referenced the removed `soft_error_summary` job output.

## Test plan

- [x] YAML parses cleanly (`yaml.safe_load`); all 11 jobs intact; 6 expected steps in the new `notify` job in the right order.
- [x] Local smoke test of `consolidate_soft_error_reports.py` against three scenarios:
  - Mixed reports (findings + `None observed` + `call_failed`) — only the real finding survives in the summary; placeholder is filtered; status pill shows `call_failed` for the broken phase.
  - All `api_skipped` reports — emits `_No soft-error findings reported by any phase._` with status pill so the operator can tell "analyser ran but found nothing" from "analyser was unreachable".
  - Empty input directory — emits the empty-consolidation stub without crashing.
- [ ] End-to-end: dispatch `Test & Mark Stable Release` on a non-release branch with `dry_run=true` (or wait for next release-gate run) to confirm the new artifact is produced and the Telegram body shows the section-aware summary.

https://claude.ai/code/session_01K1HhjbriX5LjoQoCMzsVWX

---
_Generated by [Claude Code](https://claude.ai/code/session_01K1HhjbriX5LjoQoCMzsVWX)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,6 +194,11 @@ jobs:
           set -euo pipefail
           PYTHONDONTWRITEBYTECODE=1 python3 tests/test_test_and_mark_stable_plan_polling_guard.py
 
+      - name: Soft-error report consolidator unit tests
+        run: |
+          set -euo pipefail
+          PYTHONDONTWRITEBYTECODE=1 python3 tests/test_consolidate_soft_error_reports.py
+
       - name: Workflow log collector coverage gate
         run: |
           set -euo pipefail
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,6 +199,11 @@ jobs:
           set -euo pipefail
           PYTHONDONTWRITEBYTECODE=1 python3 tests/test_consolidate_soft_error_reports.py
 
+      - name: UTF-8-safe Telegram truncation unit tests
+        run: |
+          set -euo pipefail
+          PYTHONDONTWRITEBYTECODE=1 python3 tests/test_truncate_to_utf8_byte_cap.py
+
       - name: Workflow log collector coverage gate
         run: |
           set -euo pipefail
--- a/.github/workflows/test-and-mark-stable.yml
+++ b/.github/workflows/test-and-mark-stable.yml
@@ -193,7 +193,6 @@ jobs:
       cancel-in-progress: true
     outputs:
       test_passed: ${{ steps.verify.outputs.passed }}
-      soft_error_summary: ${{ steps.soft-error-analyser.outputs.soft_error_summary }}
     env:
       GH_TOKEN: ${{ secrets.GH_PAT }}
       TEST_REPO: ${{ inputs.test_repo || github.repository }}
@@ -202,16 +201,18 @@ jobs:
       POLL_INTERVAL: 10
 
     steps:
-      # Checkout repo at the start so `if: always()` steps (Phase 8 soft-error
-      # analyser, Cleanup test artifacts) can find scripts/ even when an
-      # earlier phase exits non-zero. The original "Checkout repository for
-      # integration tests" step (Phase 5, ~line 1100) is skipped on early
-      # failure, leaving downstream `if: always()` steps with no scripts/
-      # on disk — observed in run 25115169454 where Phase 4b's bait failure
-      # cascaded into `python3: can't open file '...scripts/analyze_soft_errors.py'`
-      # and `_safe_gh_jq: command not found` in cleanup. The Phase 5 checkout
-      # is left in place (idempotent re-checkout against `ref: main`).
-      - name: Checkout repository (early — needed by Phase 8 & Cleanup on early failure)
+      # Checkout repo at the start so `if: always()` steps (per-phase
+      # soft-error analyser invocations, Cleanup test artifacts) can find
+      # scripts/ even when an earlier phase exits non-zero. The original
+      # "Checkout repository for integration tests" step (Phase 5,
+      # ~line 1100) is skipped on early failure, leaving downstream
+      # `if: always()` steps with no scripts/ on disk — observed in
+      # run 25115169454 where Phase 4b's bait failure cascaded into
+      # `python3: can't open file '...scripts/analyze_soft_errors.py'`
+      # and `_safe_gh_jq: command not found` in cleanup. The Phase 5
+      # checkout is left in place (idempotent re-checkout against
+      # `ref: main`).
+      - name: Checkout repository (early — needed by analyser & cleanup on early failure)
         uses: actions/checkout@v5
         with:
           ref: main
@@ -1665,8 +1666,9 @@ jobs:
 
             CUR_STATE="labels=${LABELS};poller=${POLLER_STATUS};conclusion=${POLLER_CONCLUSION}"
 
-            # Capture the latest poller run id (used by Phase 8's soft-error
-            # analyser to ingest the poller's logs alongside other phases).
+            # Capture the latest poller run id (used by the per-phase
+            # soft-error analyser invocation below to ingest the poller's
+            # logs as the `orchestrate_poll` phase artifact).
             if [ -n "${POLLER_RUN}" ] && [ "${POLLER_RUN}" != "null" ]; then
               POLLER_LATEST_ID=$(echo "${POLLER_RUN}" | jq -r '.id // empty')
               if [ -n "${POLLER_LATEST_ID}" ]; then
@@ -2081,118 +2083,14 @@ jobs:
           model: ${{ vars.LOG_ANALYZER_MODEL || 'openai/gpt-5.4-mini' }}
           reasoning: ${{ vars.LOG_ANALYZER_REASONING || 'medium' }}
 
-      # ── Phase 8: Soft-error analyser (non-blocking) ─────────────
-      # Pulls logs from every triggered phase run, filters to high-signal
-      # lines, and asks gpt-5.4-mini @ medium reasoning to summarise soft
-      # failures (rate-limit recoveries, codex fallbacks, summariser hard
-      # fails, etc.). The report is exposed as a job output so the notify
-      # job can attach it to the Telegram release message. Hard exits are
-      # avoided — analyser problems must never block a release.
-      # Wait briefly for every captured phase run to reach
-      # status=completed. The previous Phase 8 fired the moment Phase 7
-      # finished; on run 25126757724 that raced review_autofix's tail
-      # cleanup steps, so `gh run view --log` returned HTTP 404 ("run is
-      # still in progress") and the resulting report had no review_autofix
-      # signal at all (the phase that actually contained the root cause).
-      # The wait below caps at 5 minutes — long enough for review_autofix's
-      # cleanup tail (~40s on the failing run) and any orchestrate-poll /
-      # cancel-on-pr-close trailers to settle, short enough that a
-      # genuinely hung run does not block the release-gate Telegram
-      # notification indefinitely.
-      - name: "Phase 8a: Wait for all phase runs to complete"
-        id: phase8_wait
-        if: always()
-        continue-on-error: true
-        env:
-          GH_TOKEN: ${{ secrets.GH_PAT }}
-          CLARIFY_RUN_ID: ${{ steps.wait-clarify.outputs.run_id }}
-          PLAN_RUN_ID: ${{ steps.wait-plan.outputs.run_id }}
-          IMPL_RUN_ID: ${{ steps.wait-implement.outputs.run_id }}
-          REVIEW_RUN_ID: ${{ steps.wait-review.outputs.review_run_id }}
-          RBSIM_POLLER_RUN_ID: ${{ steps.review-blocked-sim.outputs.poller_run_id }}
-          CANCEL_RUN_ID: ${{ steps.verify-cancel-on-close.outputs.cancel_run_id }}
-        run: |
-          set -euo pipefail
-          DEADLINE=$(( $(date +%s) + 300 ))
-          for KV in \
-            "clarify:${CLARIFY_RUN_ID:-}" \
-            "plan:${PLAN_RUN_ID:-}" \
-            "implement:${IMPL_RUN_ID:-}" \
-            "review_autofix:${REVIEW_RUN_ID:-}" \
-            "orchestrate_poll:${RBSIM_POLLER_RUN_ID:-}" \
-            "cancel_on_pr_close:${CANCEL_RUN_ID:-}"; do
-            PHASE="${KV%%:*}"
-            RID="${KV#*:}"
-            [ -n "${RID}" ] && [[ "${RID}" =~ ^[0-9]+$ ]] || continue
-            while [ "$(date +%s)" -lt "${DEADLINE}" ]; do
-              STATUS=$(gh api "repos/${TEST_REPO}/actions/runs/${RID}" --jq '.status // ""' 2>/dev/null || echo "")
-              if [ "${STATUS}" = "completed" ]; then
-                echo "  ✓ ${PHASE} (run #${RID}) completed"
-                break
-              fi
-              echo "  ⏳ ${PHASE} (run #${RID}) status=${STATUS:-unknown}; waiting ..."
-              sleep 5
-            done
-          done
-
-      - name: "Phase 8: Soft-error log analyser"
-        id: soft-error-analyser
-        if: always()
-        continue-on-error: true
-        env:
-          OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
-          LOG_ANALYZER_MODEL: ${{ vars.LOG_ANALYZER_MODEL || 'openai/gpt-5.4-mini' }}
-          LOG_ANALYZER_REASONING: ${{ vars.LOG_ANALYZER_REASONING || 'medium' }}
-          CLARIFY_RUN_ID: ${{ steps.wait-clarify.outputs.run_id }}
-          PLAN_RUN_ID: ${{ steps.wait-plan.outputs.run_id }}
-          IMPL_RUN_ID: ${{ steps.wait-implement.outputs.run_id }}
-          REVIEW_RUN_ID: ${{ steps.wait-review.outputs.review_run_id }}
-          RBSIM_POLLER_RUN_ID: ${{ steps.review-blocked-sim.outputs.poller_run_id }}
-          CANCEL_RUN_ID: ${{ steps.verify-cancel-on-close.outputs.cancel_run_id }}
-        run: |
-          set -euo pipefail
-          REPORT="/tmp/soft_error_report.md"
-          ARGS=(--repo "${TEST_REPO}" --output "${REPORT}")
-          [ -n "${CLARIFY_RUN_ID:-}" ] && ARGS+=(--run "clarify=${CLARIFY_RUN_ID}")
-          [ -n "${PLAN_RUN_ID:-}" ] && ARGS+=(--run "plan=${PLAN_RUN_ID}")
-          [ -n "${IMPL_RUN_ID:-}" ] && ARGS+=(--run "implement=${IMPL_RUN_ID}")
-          [ -n "${REVIEW_RUN_ID:-}" ] && ARGS+=(--run "review_autofix=${REVIEW_RUN_ID}")
-          [ -n "${RBSIM_POLLER_RUN_ID:-}" ] && ARGS+=(--run "orchestrate_poll=${RBSIM_POLLER_RUN_ID}")
-          [ -n "${CANCEL_RUN_ID:-}" ] && ARGS+=(--run "cancel_on_pr_close=${CANCEL_RUN_ID}")
-
-          if [ "${#ARGS[@]}" -le 4 ]; then
-            echo "No phase runs captured; skipping analyser." > "${REPORT}"
-            echo "report_path=${REPORT}" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-
-          python3 scripts/analyze_soft_errors.py "${ARGS[@]}" || {
-            echo "::warning::Soft-error analyser exited non-zero — keeping fallback report"
-          }
-          echo "report_path=${REPORT}" >> "$GITHUB_OUTPUT"
-
-          # Truncate the report for downstream consumers (Telegram caps
-          # message bodies at 4096 chars; we leave headroom for the rest of
-          # the release-status text). The full untruncated report is uploaded
-          # as an artifact below.
-          if [ -f "${REPORT}" ]; then
-            head -c 2800 "${REPORT}" > /tmp/soft_error_report_short.md
-            {
-              echo 'soft_error_summary<<SOFT_ERR_EOF'
-              cat /tmp/soft_error_report_short.md
-              echo
-              echo 'SOFT_ERR_EOF'
-            } >> "$GITHUB_OUTPUT"
-          fi
-
-      - name: Upload soft-error report artifact
-        if: always() && steps.soft-error-analyser.outputs.report_path != ''
-        uses: actions/upload-artifact@v6
-        with:
-          name: soft-error-report-${{ github.run_id }}
-          path: /tmp/soft_error_report.md
-          if-no-files-found: ignore
-          retention-days: 30
+      # The previous Phase 8 in this job re-ran the LLM analyser across
+      # the 6 captured phases to produce a single combined report and
+      # truncated it byte-wise for the Telegram summary. That step was
+      # removed in favour of a deterministic concat performed in the
+      # `notify` job (see `Consolidate per-phase soft-error reports`),
+      # which can see every smoke-test sibling job's per-phase
+      # artifacts — not just this job's six. The per-phase analyser
+      # invocations above remain the source of truth.
 
       # ── Verification summary ───────────────────────────────────
       - name: Verify all phases passed
@@ -3768,8 +3666,73 @@ jobs:
     needs: [resolve-version, e2e-smoke-test, validate-scripts, orphan-workflows-test, orchestrate-decompose-test, validate-standalone-test, clarify-rejects-unsolvable-test, e2e-alt-model-test, validate, release]
     if: always()
     runs-on: ubuntu-latest
-    timeout-minutes: 2
+    timeout-minutes: 5
     steps:
+      # ── Consolidated soft-error report ─────────────────────────
+      # Replaces the historical Phase 8 LLM re-summarisation that lived
+      # inside `e2e-smoke-test`. Two reasons for moving this here:
+      #   1. Coverage: `notify` `needs:` every smoke-test job, so it can
+      #      see the per-phase artifacts emitted by orphan-workflows-test
+      #      (workflow_log_analysis, validation_refresh, update_workflows,
+      #      memory_maintenance), orchestrate-decompose-test,
+      #      validate-standalone-test, clarify-rejects-unsolvable-test,
+      #      and e2e-alt-model-test (alt_*). Phase 8 only saw the 6 e2e
+      #      phases.
+      #   2. Fidelity: pure-concat instead of an LLM second-pass means
+      #      every per-phase finding is preserved verbatim. The previous
+      #      flow re-ran the analyser model with an "<800 words" prompt
+      #      cap, which silently dropped findings during consolidation.
+      - name: Checkout repository
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 1
+
+      - name: Setup Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      - name: Download per-phase soft-error reports
+        id: download_soft_errors
+        if: always()
+        continue-on-error: true
+        uses: actions/download-artifact@v6
+        with:
+          # `actions/download-artifact@v6` accepts a glob in `pattern`
+          # and downloads matching artifacts from the *current* workflow
+          # run. Each artifact lands in its own subdirectory (default
+          # behaviour); the consolidation script tolerates both that
+          # layout and the flattened `merge-multiple` layout.
+          pattern: soft-error-report-${{ github.run_id }}-*
+          path: /tmp/soft-error-artifacts
+          merge-multiple: false
+
+      - name: Consolidate per-phase soft-error reports
+        id: consolidate_soft_errors
+        if: always()
+        continue-on-error: true
+        run: |
+          set -euo pipefail
+          mkdir -p /tmp/soft-error-artifacts
+          python3 scripts/consolidate_soft_error_reports.py \
+            --input-dir /tmp/soft-error-artifacts \
+            --output-full /tmp/soft_error_report.md \
+            --output-summary /tmp/soft_error_summary.md \
+            --max-summary-bytes 2800
+          echo "report_path=/tmp/soft_error_report.md" >> "$GITHUB_OUTPUT"
+          echo "summary_path=/tmp/soft_error_summary.md" >> "$GITHUB_OUTPUT"
+
+      - name: Upload consolidated soft-error report
+        if: always() && steps.consolidate_soft_errors.outputs.report_path != ''
+        uses: actions/upload-artifact@v6
+        with:
+          # Artifact name preserved for backward compatibility with any
+          # operator tooling that fetches `soft-error-report-<run_id>`.
+          name: soft-error-report-${{ github.run_id }}
+          path: /tmp/soft_error_report.md
+          if-no-files-found: ignore
+          retention-days: 30
+
       - name: Send Telegram notification
         id: tg_send
         env:
@@ -3833,11 +3796,20 @@ jobs:
           fi
           MSG+=$'\n'"Run: ${RUN_URL}"
 
-          # Append the soft-error analyser report (truncated upstream to fit
-          # comfortably within Telegram's 4096-char body cap). Non-blocking
-          # by design: any soft-failure findings are surfaced here so the
-          # release engineer sees them inline with the release notification.
-          SOFT_ERR='${{ needs.e2e-smoke-test.outputs.soft_error_summary }}'
+          # Append the consolidated soft-error summary built by the
+          # `Consolidate per-phase soft-error reports` step above. The
+          # script does section-aware truncation (it drops whole findings
+          # blocks rather than slicing mid-finding) and emits a status
+          # pill listing every phase's analyser status code so the
+          # release engineer can spot analyser failures (call_failed /
+          # api_skipped) at a glance even when there are no findings.
+          # Non-blocking by design: a missing summary file means the
+          # consolidation step itself failed, in which case the
+          # release-status header is still delivered.
+          SOFT_ERR=""
+          if [ -f /tmp/soft_error_summary.md ]; then
+            SOFT_ERR=$(cat /tmp/soft_error_summary.md)
+          fi
           if [ -n "${SOFT_ERR}" ]; then
             MSG+=$'\n\n'"---"$'\n'"${SOFT_ERR}"
           fi
--- a/.github/workflows/test-and-mark-stable.yml
+++ b/.github/workflows/test-and-mark-stable.yml
@@ -3792,17 +3792,26 @@ jobs:
 
           # Sanitize SOFT_ERROR_TELEGRAM_BYTES in shell before passing
           # to Python. `set -euo pipefail` is in effect, so a non-integer
-          # repo-var value (empty string, whitespace, "abc", etc.) would
-          # otherwise make argparse exit non-zero and abort the step
-          # before it writes report_path/summary_path — undermining the
-          # documented fail-open intent. Coerce invalid input to the
-          # default (2800) and emit a workflow warning so misconfiguration
-          # is visible without breaking the release-status notification.
+          # repo-var value (whitespace, "abc", etc.) would otherwise make
+          # argparse exit non-zero and abort the step before it writes
+          # report_path/summary_path — undermining the documented
+          # fail-open intent.
+          #
+          # Trimming first means a repo-var saved as `" 3000 "` (with
+          # incidental whitespace) is treated equivalently to `"3000"`
+          # rather than silently falling back to the default. The
+          # subsequent regex accepts any non-negative integer including
+          # `0`, which the Python consolidator interprets as "disable
+          # truncation" via its `max_bytes <= 0` branch — operators
+          # who set `SOFT_ERROR_TELEGRAM_BYTES=0` to opt out of the
+          # cap have their intent honoured rather than overridden.
           SOFT_ERROR_BYTES_RAW="${SOFT_ERROR_TELEGRAM_BYTES:-2800}"
-          if [[ "${SOFT_ERROR_BYTES_RAW}" =~ ^[0-9]+$ ]] && [ "${SOFT_ERROR_BYTES_RAW}" -gt 0 ]; then
-            SOFT_ERROR_BYTES="${SOFT_ERROR_BYTES_RAW}"
+          SOFT_ERROR_BYTES_TRIMMED="${SOFT_ERROR_BYTES_RAW#"${SOFT_ERROR_BYTES_RAW%%[![:space:]]*}"}"
+          SOFT_ERROR_BYTES_TRIMMED="${SOFT_ERROR_BYTES_TRIMMED%"${SOFT_ERROR_BYTES_TRIMMED##*[![:space:]]}"}"
+          if [[ "${SOFT_ERROR_BYTES_TRIMMED}" =~ ^[0-9]+$ ]]; then
+            SOFT_ERROR_BYTES="${SOFT_ERROR_BYTES_TRIMMED}"
           else
-            echo "::warning::Invalid SOFT_ERROR_TELEGRAM_BYTES='${SOFT_ERROR_BYTES_RAW}'; falling back to 2800."
+            echo "::warning::Invalid SOFT_ERROR_TELEGRAM_BYTES='${SOFT_ERROR_BYTES_RAW}' (must be a non-negative integer); falling back to 2800."
             SOFT_ERROR_BYTES=2800
           fi
 
--- a/.github/workflows/test-and-mark-stable.yml
+++ b/.github/workflows/test-and-mark-stable.yml
@@ -3682,12 +3682,22 @@ jobs:
       #      every per-phase finding is preserved verbatim. The previous
       #      flow re-ran the analyser model with an "<800 words" prompt
       #      cap, which silently dropped findings during consolidation.
+      # The notify job runs `if: always()` to deliver a Telegram message
+      # even when earlier jobs fail. Each setup step below is marked
+      # fail-open so a transient checkout/setup-python/download outage
+      # cannot suppress the release-status notification — the
+      # consolidation step degrades gracefully (empty-input stub) and
+      # the Telegram send step itself is `if: always()`.
       - name: Checkout repository
+        if: always()
+        continue-on-error: true
         uses: actions/checkout@v5
         with:
           fetch-depth: 1
 
       - name: Setup Python
+        if: always()
+        continue-on-error: true
         uses: actions/setup-python@v6
         with:
           python-version: "3.12"
@@ -3735,6 +3745,13 @@ jobs:
 
       - name: Send Telegram notification
         id: tg_send
+        # Always send the release-status notification, even if any of the
+        # upstream consolidation steps (checkout / setup-python / download
+        # / consolidate) failed. Without `if: always()` a single failure
+        # in those steps would suppress the notification entirely; the
+        # Telegram body simply omits the soft-error summary when its
+        # source file is missing.
+        if: always()
         env:
           TG_BOT_SECRET: ${{ secrets.TG_BOT_SECRET }}
           TG_CHAT_ID: ${{ vars.TG_ADMIN_CHAT_ID }}
--- a/.github/workflows/test-and-mark-stable.yml
+++ b/.github/workflows/test-and-mark-stable.yml
@@ -3721,6 +3721,14 @@ jobs:
         id: consolidate_soft_errors
         if: always()
         continue-on-error: true
+        env:
+          # Surface the download step's outcome so the consolidator can
+          # emit a clearly-marked "download failed" stub if the input
+          # directory is empty for that reason — without this, a
+          # transient download outage would silently produce a "no
+          # findings" stub identical to a clean run, masking real
+          # per-phase findings that exist upstream.
+          DOWNLOAD_OUTCOME: ${{ steps.download_soft_errors.outcome }}
         run: |
           set -euo pipefail
           mkdir -p /tmp/soft-error-artifacts
@@ -3728,7 +3736,8 @@ jobs:
             --input-dir /tmp/soft-error-artifacts \
             --output-full /tmp/soft_error_report.md \
             --output-summary /tmp/soft_error_summary.md \
-            --max-summary-bytes 2800
+            --max-summary-bytes 2800 \
+            --download-status "${DOWNLOAD_OUTCOME:-unknown}"
           echo "report_path=/tmp/soft_error_report.md" >> "$GITHUB_OUTPUT"
           echo "summary_path=/tmp/soft_error_summary.md" >> "$GITHUB_OUTPUT"
 
--- a/.github/workflows/test-and-mark-stable.yml
+++ b/.github/workflows/test-and-mark-stable.yml
@@ -3689,6 +3689,7 @@ jobs:
       # consolidation step degrades gracefully (empty-input stub) and
       # the Telegram send step itself is `if: always()`.
       - name: Checkout repository
+        id: checkout_repo
         if: always()
         continue-on-error: true
         uses: actions/checkout@v5
@@ -3696,6 +3697,7 @@ jobs:
           fetch-depth: 1
 
       - name: Setup Python
+        id: setup_python
         if: always()
         continue-on-error: true
         uses: actions/setup-python@v6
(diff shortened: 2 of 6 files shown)


shubhodeep1:
## Claude-branch review

* Commit: `45dca567fc2d4165d527f3e27fd63124ab5b2c46`
* Ref: `claude/test-error-reporting-cLOIx`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25155963108

=== CONSENSUS FINDINGS ===
- scripts/consolidate_soft_error_reports.py:365-365 | severity=low | confidence=2
  flagged_by: [qwen_qwen3_6-plus]
  PROBLEM: `download_failed = args.download_status not in ("", "success")` can emit a misleading "DOWNLOAD FAILED" stub when the download-status plumbing is missing or miswired.
  WHY: qwen notes this is a low-confidence false-alarm path driven by treating any non-empty non-success token as failure, and suggests enumerating the actual failure outcomes instead.

(No findings reported.)

(No findings reported.)

(No findings reported.)

- scripts/consolidate_soft_error_reports.py:365 | confidence=2
  PROBLEM: `download_failed = args.download_status not in ("", "success")`
  WHY: When `${DOWNLOAD_OUTCOME:-unknown}` falls back to `"unknown"` or the env plumbing breaks, the consolidator may emit a "DOWNLOAD FAILED" stub that is indistinguishable from a real download outage.

(No findings reported.)

(No findings reported.)

shubhodeep1:
## Claude-branch review

* Commit: `7ebc582e1eada2b3b5439843f9f6c0565b5851c2`
* Ref: `claude/test-error-reporting-cLOIx`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25157829636

=== CONSENSUS FINDINGS ===
- .github/workflows/test-and-mark-stable.yml:3706-3729 | severity=low | confidence=5
  flagged_by: [moonshotai_kimi-k2_5, x-ai_grok-4_1-fast]
  PROBLEM: The fallback stub tells operators to inspect per-phase artifacts even though the fallback path is specifically for missing checkout/python/script prerequisites, which can make the guidance misleading.
  WHY: Both reviewers flagged the same shell fallback block as confusing about artifact availability; moonshot called it a clarity issue and x-ai said it can send operators chasing artifacts instead of the real failure.
- .github/workflows/test-and-mark-stable.yml:3720-3751 | severity=low | confidence=4
  flagged_by: [minimax_minimax-m2_5, moonshotai_kimi-k2_5, x-ai_grok-4_1-fast]
  PROBLEM: The workflow hardcodes `--max-summary-bytes 2800` instead of exposing the summary cap as a tunable variable.
  WHY: All three reviewers identified the same hardcoded limit as an operational flexibility issue rather than a runtime bug.

(No findings reported.)

- .github/workflows/test-and-mark-stable.yml:3721 | severity=low | confidence=3
  PROBLEM: The `--max-summary-bytes 2800` limit is hardcoded in the workflow instead of being configurable at runtime.
  WHY: The reviewer called this a minor hardening gap because operators cannot tune the cap without editing the workflow file.

- .github/workflows/test-and-mark-stable.yml:3723-3728 | severity=low | confidence=3
  PROBLEM: The fallback stub points operators at per-phase artifacts in a way that can be confusing on the fallback path.
  WHY: The reviewer said the stub correctly identifies the primary failure, but the remediation suggestion could mislead if the download step also failed.
- .github/workflows/test-and-mark-stable.yml:3751 | severity=low | confidence=2
  PROBLEM: The `--max-summary-bytes 2800` limit is hardcoded in the workflow and cannot be tuned without modifying the file.
  WHY: The reviewer treated this as a minor operational limitation rather than a correctness issue.

(No findings reported.)

- .github/workflows/test-and-mark-stable.yml:3706-3729 | severity=low | confidence=5
  PROBLEM: The fallback stub references per-phase soft-error artifacts as potentially inspectable even though that fallback path is about missing local prerequisites, which can send operators toward the wrong diagnosis.
  WHY: The reviewer said the stub can make an operator chase artifact list entries instead of the actual checkout/setup failure.
- .github/workflows/test-and-mark-stable.yml:3720-3721 | severity=low | confidence=4
  PROBLEM: The workflow hardcodes `--max-summary-bytes 2800` with no variable or input driving the cap.
  WHY: The reviewer flagged this as a configurability gap that limits tuning for Telegram payload variance.

(No findings reported.)

shubhodeep1:
## Claude-branch review

* Commit: `cec03c764a09f65e184be0e3a02b457c0eebf6ea`
* Ref: `claude/test-error-reporting-cLOIx`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25159596784

=== CONSENSUS FINDINGS ===
- .github/workflows/test-and-mark-stable.yml:3800-3807 | severity=critical | confidence=5
  flagged_by: [deepseek_deepseek-v4-pro, minimax_minimax-m2_5, moonshotai_kimi-k2_5, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The shell sanitization rejects `0`, so `SOFT_ERROR_TELEGRAM_BYTES=0` falls back to `2800` instead of disabling truncation as the Python helper allows.
  WHY: This contradicts the Python-side `cap <= 0` no-truncation behavior and overrides an explicit operator setting; severity is critical because one reviewer labeled it critical while others reported medium/high or omitted severity.
- .github/workflows/test-and-mark-stable.yml:3798-3810 | severity=critical | confidence=5
  flagged_by: [deepseek_deepseek-v4-pro, minimax_minimax-m2_5, moonshotai_kimi-k2_5, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The shell regex rejects whitespace-padded integers, so valid numeric caps with incidental spaces fall back to `2800`.
  WHY: A repo-var value containing leading or trailing whitespace is treated as invalid and replaced with the default, silently discarding operator intent; severity is critical per the strongest reviewer and others reported medium/high or no severity.
- scripts/consolidate_soft_error_reports.py:157-177 | severity=low | confidence=3
  flagged_by: [minimax_minimax-m2_5, moonshotai_kimi-k2_5, x-ai_grok-4_1-fast]
  PROBLEM: The truncation helper decodes with `errors="ignore"`, so malformed UTF-8 bytes are silently dropped without any diagnostic.
  WHY: This can mask upstream corruption in rare malformed-input cases; one reviewer considered the behavior acceptable, but the others flagged the silent-drop path as a bug.

- .github/workflows/test-and-mark-stable.yml:3801-3807 | confidence=4
  PROBLEM: The shell sanitization block rejects `0` as invalid and falls back to `2800`.
  WHY: The condition `[ "${SOFT_ERROR_BYTES_RAW}" -gt 0 ]` explicitly rejects `0`, but the Python code treats `max_bytes <= 0` as no truncation, so an operator setting `SOFT_ERROR_TELEGRAM_BYTES=0` gets unexpected truncation.
- .github/workflows/test-and-mark-stable.yml:3802 | confidence=2
  PROBLEM: The bash regex does not handle whitespace-padded integers.
  WHY: `^[0-9]+$` rejects values like `" 3000 "`, so the sanitization falls through and silently replaces the intended cap with `2800`.

- .github/workflows/test-and-mark-stable.yml:~3801 | confidence=4
  PROBLEM: Workflow sanitization rejects `0` as invalid.
  WHY: The `-gt 0` check rejects `0`, but `0` is a valid value that could mean disable truncation, so the operator's explicit intent is silently overridden.
- .github/workflows/test-and-mark-stable.yml:~3801 | confidence=4
  PROBLEM: Workflow sanitization rejects whitespace-padded integers.
  WHY: `^[0-9]+$` requires the entire string to be digits, so values like `" 3000 "` or `"3000 "` fail validation and fall back to `2800`.
- scripts/consolidate_soft_error_reports.py:~177 | confidence=3
  PROBLEM: The truncation helper silently drops malformed trailing UTF-8 bytes with `errors="ignore"`.
  WHY: If artifact bytes are corrupted, the summary can be mangled without any diagnostic, obscuring upstream data corruption.

- .github/workflows/test-and-mark-stable.yml:3801 | severity=critical | confidence=5
  PROBLEM: The validation rejects `0` and whitespace-padded integers as invalid, forcing them to fall back to `2800`.
  WHY: `0` is a valid operator intent meaning disable truncation, but it is rejected as invalid, and the regex also lacks whitespace tolerance for values like `" 3000 "`.
- scripts/consolidate_soft_error_reports.py:176 | severity=minor | confidence=3
  PROBLEM: `decode("utf-8", errors="ignore")` silently discards malformed UTF-8 bytes without diagnostic.
  WHY: In the degenerate case where bytes become corrupted during processing, malformed sequences are silently dropped rather than replaced visibly, even though `_read_report_safe` uses `errors="replace"` for visibility.

(No findings reported.)

- .github/workflows/test-and-mark-stable.yml:3798-3810 | confidence=5
  PROBLEM: `[[ "${SOFT_ERROR_BYTES_RAW}" =~ ^[0-9]+$ ]]` rejects whitespace-padded integers like `" 3000 "` or `"3000 "`.
  WHY: Valid repo-var values with inadvertent whitespace are treated as invalid, forcing fallback to 2800 and potentially truncating summaries longer than intended.
- .github/workflows/test-and-mark-stable.yml:3801-3807 | confidence=5
  PROBLEM: `[ "${SOFT_ERROR_BYTES_RAW}" -gt 0 ]` explicitly rejects `0`, forcing fallback to 2800.
  WHY: An operator cannot disable truncation via `SOFT_ERROR_TELEGRAM_BYTES=0`, so summaries are forcibly capped even when full text is desired.
- scripts/consolidate_soft_error_reports.py:162-174 | confidence=3
  PROBLEM: `encoded[:i].decode("utf-8", errors="ignore")` silently drops malformed trailing bytes post-truncation.
  WHY: Corrupted artifacts can produce mangled summaries without a diagnostic, making it harder to distinguish truncation from upstream corruption.
- scripts/consolidate_soft_error_reports.py:157-176 | confidence=3
  PROBLEM: Existing `decode("utf-8", errors="ignore")` silently drops invalid UTF-8 sequences.
  WHY: Non-UTF8 artifact content can mangle findings extraction without warning.

- .github/workflows/test-and-mark-stable.yml:3800-3806 | confidence=5
  PROBLEM: The bash sanitization logic rejects `0` as invalid and falls back to `2800`.
  WHY: `0` is explicitly supported by the Python consolidator as disable/no truncation, so the workflow silently ignores an operator's documented setting.
- .github/workflows/test-and-mark-stable.yml:3801 | confidence=4
  PROBLEM: The regex `^[0-9]+$` does not handle whitespace-padded values like `" 3000 "`.
  WHY: A repo var with accidental leading or trailing spaces fails validation and causes the workflow to use `2800` instead of the intended value.

shubhodeep1:
## Claude-branch review

* Commit: `f3b2546a11dfe5d222930e2e89605f9a3104cf6d`
* Ref: `claude/test-error-reporting-cLOIx`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25161156595

=== CONSENSUS FINDINGS ===
- scripts/truncate_to_utf8_byte_cap.py:37-52 | confidence=5
  PROBLEM: The walk-back logic drops a complete multi-byte codepoint when the byte cap lands exactly on a codepoint boundary.
  WHY: Multiple reviewers reproduced shorter-than-necessary valid UTF-8 output at exact boundaries, so the truncator silently discards bytes that already fit.
- scripts/truncate_to_utf8_byte_cap.py:42-52 | confidence=4
  flagged_by: [minimax_minimax-m2_5, moonshotai_kimi-k2_5, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The leader-byte check treats invalid UTF-8 starter bytes as valid leaders.
  WHY: Bytes 0xC0-0xC1 and 0xF5-0xFF are not legal UTF-8 starts, so malformed input can be misclassified and handled incorrectly.
- .github/workflows/test-and-mark-stable.yml:3928-3933 | confidence=5
  flagged_by: [minimax_minimax-m2_5, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The bash fallback `${MSG:0:3950}` truncates by characters rather than bytes.
  WHY: In the no-python path, multi-byte UTF-8 can exceed Telegram's byte limit or be split mid-codepoint.
- .github/workflows/test-and-mark-stable.yml:3950-3953 | confidence=4
  flagged_by: [deepseek_deepseek-v4-pro, moonshotai_kimi-k2_5]
  PROBLEM: The fallback slice can exceed Telegram's byte limit because it counts characters, not bytes.
  WHY: This degraded path can produce an oversized message when the payload contains many multi-byte characters.
- .github/workflows/test-and-mark-stable.yml:3924 | confidence=3
  PROBLEM: `MSG_BYTES=$(printf '%s' "${MSG}" | wc -c)` can undercount messages containing embedded NUL bytes.
  WHY: `printf '%s'` stops at the first NUL, so the byte-length guard can skip truncation for an oversized payload.
- .github/workflows/test-and-mark-stable.yml:3949 | confidence=2
  flagged_by: [deepseek_deepseek-v4-pro]
  PROBLEM: `MSG_BYTES=$(printf '%s' "${MSG}" | wc -c)` can undercount messages containing embedded NUL bytes.
  WHY: The same `%s` behavior hides bytes after a NUL, allowing an oversized message to reach Telegram.
- tests/test_truncate_to_utf8_byte_cap.py:46-60 | confidence=4
  PROBLEM: The Unicode truncation test checks only length and UTF-8 validity, not that the output is the maximal valid prefix.
  WHY: The buggy helper still produces a shorter-but-valid result, so the test passes while the data-loss bug remains hidden.
- scripts/truncate_to_utf8_byte_cap.py:1 | confidence=2
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: The new script is not marked executable, so direct invocation with `./` fails.
  WHY: Without +x mode, manual or future workflow calls that execute the script directly will error with ENOEXEC.
- scripts/consolidate_soft_error_reports.py:159-182 | confidence=3
  flagged_by: [z-ai_glm-5]
  PROBLEM: The consolidator uses different UTF-8 truncation logic than the standalone helper.
  WHY: The two code paths do not share the same boundary handling, so the tested helper and production consolidator can diverge.

- scripts/truncate_to_utf8_byte_cap.py:43-49 | confidence=5
  PROBLEM: The walk-back logic removes a complete multi-byte character when the byte cap lands exactly on a codepoint boundary.
  WHY: The loop backs over continuation bytes and then unconditionally drops the preceding leader, which makes exact-boundary slices shorter than necessary.
- tests/test_truncate_to_utf8_byte_cap.py:46-53 | confidence=4
  PROBLEM: The Unicode truncation test does not check that the output is the longest valid prefix.
  WHY: A shorter-but-valid output still satisfies the current length and decode assertions, so the boundary bug can pass CI.
- .github/workflows/test-and-mark-stable.yml:3950 | confidence=3
  PROBLEM: The fallback truncation `${MSG:0:3950}` counts characters instead of bytes.
  WHY: In the degraded no-python path, multi-byte UTF-8 can exceed Telegram's byte limit or be cut mid-codepoint.
- .github/workflows/test-and-mark-stable.yml:3949 | confidence=2
  PROBLEM: `MSG_BYTES=$(printf '%s' "${MSG}" | wc -c)` can undercount when MSG contains embedded NUL bytes.
  WHY: `printf '%s'` stops at the first NUL, so the guard may miss an oversized payload.

- scripts/truncate_to_utf8_byte_cap.py:37-52 | confidence=5
  PROBLEM: The orphan-leader drop logic removes complete multi-byte characters when the cap lands exactly on a codepoint boundary.
  WHY: When the slice already ends on a valid boundary, the unconditional leader drop makes the output shorter than the maximal valid prefix.
- scripts/truncate_to_utf8_byte_cap.py:42-47 | confidence=3
  PROBLEM: The leader-byte check is too permissive and treats invalid UTF-8 starters as valid leaders.
  WHY: Bytes 0xC0-0xC1 and 0xF5-0xFF are not valid UTF-8 starts, so malformed input can be misclassified.
- .github/workflows/test-and-mark-stable.yml:3928-3933 | confidence=5
  PROBLEM: The bash fallback `${MSG:0:3950}` slices by characters, not bytes.
  WHY: When Python is unavailable, multi-byte UTF-8 can exceed the intended byte cap or be split mid-codepoint.
- .github/workflows/test-and-mark-stable.yml:3924 | confidence=3
  PROBLEM: `MSG_BYTES=$(printf '%s' "${MSG}" | wc -c)` can undercount if MSG contains NUL bytes.
  WHY: `printf '%s'` stops at the first NUL, so the byte guard can be skipped for an oversized message.

- scripts/truncate_to_utf8_byte_cap.py:43-47 | confidence=5
  PROBLEM: The orphan-leader logic drops complete multi-byte characters at exact byte boundaries.
  WHY: The code walks back over continuations and then always removes the preceding leader, even when the full codepoint already fits.
- .github/workflows/test-and-mark-stable.yml:3953 | confidence=4
  PROBLEM: The bash fallback truncation uses character slicing instead of byte-aware truncation.
  WHY: In the fallback path, `${MSG:0:3950}` can produce an oversized or corrupted UTF-8 message.
- tests/test_truncate_to_utf8_byte_cap.py:52-60 | confidence=4
  PROBLEM: The Unicode truncation test does not verify maximality.
  WHY: It checks only `len(out) <= cap` and UTF-8 validity, so an undersized but valid result still passes.
- scripts/truncate_to_utf8_byte_cap.py:42-47 | confidence=2
  PROBLEM: The leader-byte check accepts invalid UTF-8 starter ranges.
  WHY: Bytes 0xC0-0xC1 and 0xF5-0xFF are not legal UTF-8 starts, so malformed input is handled too loosely.

- scripts/truncate_to_utf8_byte_cap.py:39-47 | confidence=5
  PROBLEM: The walk-back logic treats a valid codepoint leader at the boundary as an orphan leader and drops it.
  WHY: Exact-boundary slices can lose a complete multi-byte glyph, making the output shorter than the maximal valid prefix.

- scripts/truncate_to_utf8_byte_cap.py:37-52 | confidence=5
  PROBLEM: The walk-back logic drops a full multi-byte codepoint when the cap lands exactly on a codepoint boundary.
  WHY: The slice is already valid at the boundary, so removing the leader makes the result shorter than necessary.
- scripts/truncate_to_utf8_byte_cap.py:42-47 | confidence=4
  PROBLEM: The leader-byte check accepts invalid UTF-8 starters.
  WHY: Bytes 0xC0-0xC1 and 0xF5-0xFF are not legal UTF-8 starts, so malformed input can be mishandled.
- tests/test_truncate_to_utf8_byte_cap.py:52-60 (test_truncate_with_emoji_no_mid_byte_split) | confidence=4
  PROBLEM: The test checks only length and successful decode, not maximality.
  WHY: The exact-boundary undershoot bug still passes because the shorter output remains valid UTF-8.
- .github/workflows/test-and-mark-stable.yml:3924 | confidence=3
  PROBLEM: `MSG_BYTES=$(printf '%s' "${MSG}" | wc -c)` undercounts if MSG contains embedded NUL bytes.
  WHY: The `%s` format stops at the first NUL, so the truncation guard can miss an oversized payload.
- scripts/truncate_to_utf8_byte_cap.py:1 | confidence=2
  PROBLEM: The new script is missing executable permission.
  WHY: Direct `./` invocation fails without +x mode, even though the workflow currently calls it through Python.

- scripts/truncate_to_utf8_byte_cap.py:42-52 | confidence=5
  PROBLEM: The UTF-8 boundary walk-back logic produces undersized output when the cap lands exactly on a codepoint boundary.
  WHY: The algorithm walks back from the cap and then unconditionally drops the leader byte, which removes characters that already fit.
- scripts/truncate_to_utf8_byte_cap.py:42-47 | confidence=3
  PROBLEM: The leader-byte check includes invalid UTF-8 starter ranges.
  WHY: Bytes 0xC0-0xC1 and 0xF5-0xFF are not valid UTF-8 leaders, so malformed input can be misclassified.
- .github/workflows/test-and-mark-stable.yml:3928-3933 (fallback path `MSG="${MSG:0:3950}"`) | confidence=5
  PROBLEM: The bash fallback truncation counts characters rather than bytes.
  WHY: Multi-byte UTF-8 in the fallback path can exceed the intended byte limit or be split mid-codepoint.
- tests/test_truncate_to_utf8_byte_cap.py:52-60 | confidence=4
  PROBLEM: The Unicode truncation test does not verify the maximal valid prefix.
  WHY: A shorter-but-valid result still satisfies the current assertions, so the data-loss bug can pass unnoticed.
- scripts/consolidate_soft_error_reports.py:159-182 (`_truncate_bytes_at_utf8_boundary`) | confidence=3
  PROBLEM: The consolidator uses a different truncation algorithm than the standalone helper.
  WHY: The two code paths diverge on boundary handling, so the tested helper and production consolidator can disagree.

shubhodeep1:
## Claude-branch review

* Commit: `12714b85f6347ae00acae5b77060c1dd9c525e23`
* Ref: `claude/test-error-reporting-cLOIx`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25163198735

=== CONSENSUS FINDINGS ===
- scripts/consolidate_soft_error_reports.py:169-187 | severity=medium | confidence=3
  flagged_by: [qwen_qwen3_6-plus]
  PROBLEM: Removing the `cap <= 0` early return changes `_truncate_bytes_at_utf8_boundary` so `cap=0` now returns the full input instead of `""`.
  WHY: The reviewer says this is a semantic regression for future callers that relied on the old empty-string-on-zero-cap behavior.
- scripts/consolidate_soft_error_reports.py:312-330 | severity=medium | confidence=3
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: The truncation loop re-encodes the growing prefix each iteration, which can make the `build_summary` path O(n²) for large finding lists.
  WHY: The reviewer says repeated `encode("utf-8")` calls on the accumulating string scale poorly with many blocks.
- scripts/consolidate_soft_error_reports.py:333-340 | severity=low | confidence=4
  flagged_by: [minimax_minimax-m2_5, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: Inline truncation assembly inserts an extra newline, creating a double blank line before the truncated finding content.
  WHY: Reviewers point out that `head` already ends with `\n`, so adding another `"\n"` before `inline` yields `\n\n`; they describe it as a cosmetic formatting glitch.
- scripts/consolidate_soft_error_reports.py:335 | severity=low | confidence=2
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: Accessing `finding_blocks[0]` is tightly coupled to the nearby guards and could become unsafe if the guard order changes.
  WHY: The reviewer says it is safe today because `budget > 0 and finding_blocks` short-circuits first, but the coupling is fragile.
- scripts/consolidate_soft_error_reports.py:351-354 | severity=low | confidence=2
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: The last-resort truncation guard depends on `_truncate_bytes_at_utf8_boundary` to enforce the byte cap after assembly overflow.
  WHY: The reviewer says the guard works, but it only fires after head+marker overflow and is an edge-case safety path.
- scripts/truncate_to_utf8_byte_cap.py:28-72 | severity=low | confidence=2
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: `truncate_bytes_to_utf8_cap` assumes valid UTF-8 input and does not validate malformed leading-byte sequences itself.
  WHY: The reviewer says the helper relies on its contract and corrupted artifacts would surface as replacement glyphs in the caller.
- .github/workflows/test-and-mark-stable.yml:3934-3964 | severity=medium | confidence=3
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: The fallback truncation path uses `head -c 3950`, which can split a UTF-8 codepoint and garble the tail.
  WHY: The reviewer says the fallback is byte-oriented and can emit U+FFFD or otherwise garbled output when truncating mid-codepoint.
- .github/workflows/test-and-mark-stable.yml:3947 | severity=medium | confidence=3
  PROBLEM: `printf '%s' "${MSG}" | wc -c` can undercount if `MSG` contains NUL bytes.
  WHY: One reviewer called it a theoretical edge case and another said it could skip truncation and exceed Telegram's cap.
- scripts/truncate_to_utf8_byte_cap.py:68 | severity=low | confidence=3
  flagged_by: [minimax_minimax-m2_5]
  PROBLEM: The reviewer reports a docstring/implementation mismatch around the `cap=0` behavior.
  WHY: They say the docstring claimed one behavior while the code returns the input unchanged, which could confuse maintainers.
- scripts/truncate_to_utf8_byte_cap.py:27-71 | severity=low | confidence=2
  flagged_by: [moonshotai_kimi-k2_5]
  PROBLEM: The helper returns an unexpected tail slice when called with `cap < 0`.
  WHY: The reviewer says negative caps are not currently reachable from production code, but they make the function contract fragile.

(No findings reported.)

- scripts/consolidate_soft_error_reports.py:339-340 | severity=low | confidence=4
  PROBLEM: Double newline inserted when assembling inline truncation fallback.
  WHY: The code constructs `head` which already ends with `\n`, then adds another `"\n"` when concatenating, producing `\n\n` between head and inline content.
- .github/workflows/test-and-mark-stable.yml:3947 | severity=medium | confidence=3
  PROBLEM: Byte count can undercount when NUL bytes are present in the message.
  WHY: The code uses `printf '%s' "${MSG}" | wc -c`; if a corrupted artifact contained NUL bytes, the count could be incorrect and truncation could be skipped.
- scripts/truncate_to_utf8_byte_cap.py:68 | severity=low | confidence=3
  PROBLEM: The reviewer reports a docstring/implementation mismatch around the `cap=0` behavior.
  WHY: They say the docstring says `returns "" when cap <= 0`, but the implementation returns the input unchanged, which could confuse maintainers.

- scripts/truncate_to_utf8_byte_cap.py:27-71 | severity=low | confidence=2
  PROBLEM: The helper returns an unexpected tail slice when called with `cap < 0`.
  WHY: The reviewer says negative caps are not currently reachable from production code, but they make the function contract fragile.

- scripts/consolidate_soft_error_reports.py:169-187 | severity=medium | confidence=3
  PROBLEM: Removing the `cap <= 0` early return changes `_truncate_bytes_at_utf8_boundary` so `cap=0` now returns the full input instead of `""`.
  WHY: The reviewer says this is a semantic regression for any future caller that relied on the old empty-string-on-zero-cap behavior.

- scripts/consolidate_soft_error_reports.py:312-330 | severity=medium | confidence=3
  PROBLEM: Incremental byte tracking re-encodes the running string each iteration, making the truncation loop potentially O(n²).
  WHY: The reviewer says repeated `encode("utf-8")` calls on the accumulating prefix scale poorly for large `finding_blocks` lists.
- scripts/consolidate_soft_error_reports.py:335 | severity=low | confidence=2
  PROBLEM: `finding_blocks[0]` is accessed under a guard that is tightly coupled to the current branch order.
  WHY: The reviewer says it is safe today because `budget > 0 and finding_blocks` short-circuits first, but the coupling is fragile if guards are reordered.
- scripts/consolidate_soft_error_reports.py:339-340 | severity=low | confidence=3
  PROBLEM: The inline truncation assembly adds an extra newline, producing a double newline before the inline block.
  WHY: The reviewer says `head` already ends with `\n`, so adding another `"\n"` before `inline` creates `\n\n`.
- scripts/consolidate_soft_error_reports.py:351-354 | severity=low | confidence=2
  PROBLEM: The last-resort cap guard only triggers after assembly overflow and depends on the helper to enforce the byte limit.
  WHY: The reviewer says the guard works, but it is an edge-case safety path rather than a direct guarantee.
- scripts/truncate_to_utf8_byte_cap.py:28-72 | severity=low | confidence=2
  PROBLEM: The helper assumes valid UTF-8 input and does not validate malformed leading-byte sequences itself.
  WHY: The reviewer says corrupted artifacts would surface as replacement glyphs in the caller and the helper relies on its contract.
- .github/workflows/test-and-mark-stable.yml:3934-3964 | severity=medium | confidence=3
  PROBLEM: The fallback truncation path uses `head -c 3950`, which can split a UTF-8 codepoint and garble the tail.
  WHY: The reviewer says the fallback is byte-oriented and can emit U+FFFD or otherwise garbled output when truncating mid-codepoint.
- .github/workflows/test-and-mark-stable.yml:3947 | severity=medium | confidence=3
  PROBLEM: `printf '%s' "${MSG}" | wc -c` can undercount if `MSG` contains NUL bytes.
  WHY: The reviewer says this can skip truncation and exceed Telegram's cap.

- scripts/consolidate_soft_error_reports.py:333 | severity=low | confidence=3
  PROBLEM: When `head` already ends with `\n`, adding another `"\n"` before `inline` creates a double newline between the status pill and the truncated finding content.
  WHY: The reviewer says this is a cosmetic formatting inconsistency in the inline truncation path.

shubhodeep1:
## Claude-branch review

* Commit: `50b0513f26999289dcc80e0c7685ffa56b508948`
* Ref: `claude/test-error-reporting-cLOIx`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25165138809

=== CONSENSUS FINDINGS ===
- tests/test_consolidate_soft_error_reports.py:111-140 | severity=medium | confidence=5
  PROBLEM: The monkeypatch-based layout-1 OSError test does not reliably validate the intended layout-2 fallback path.
  WHY: Reviewers disagreed on whether `glob()` is also hit or remains unpatched, but they agreed the current monkeypatch does not prove the documented recovery behavior.
- tests/test_consolidate_soft_error_reports.py:141-152 | severity=low | confidence=3
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: The `test_discover_oserror_on_is_dir_returns_empty` case only patches the root `is_dir()` call, so it does not prove the per-child `is_dir()` path stays fail-open.
  WHY: This exercises a different error site than the one the surrounding code is meant to harden.
- .github/workflows/test-and-mark-stable.yml:3771-3813 | severity=low | confidence=3
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: The fallback summary block describes the step as a crash even though the `||` branch can run for any non-zero Python exit.
  WHY: That makes the step summary misleading when the Python step fails for reasons other than an actual crash.
- .github/workflows/test-and-mark-stable.yml:3815-3821 | severity=low | confidence=3
  flagged_by: [deepseek_deepseek-v4-pro, minimax_minimax-m2_5, moonshotai_kimi-k2_5, qwen_qwen3_6-plus, z-ai_glm-5]
  PROBLEM: The shell sanitizer rejects `SOFT_ERROR_TELEGRAM_BYTES=-1` even though the Python consolidator treats `max_bytes <= 0` as truncation disabled.
  WHY: Reviewers disagreed on whether only `0` should be supported, but they all noted the shell gate and Python behavior use different sentinel rules.
- .github/workflows/test-and-mark-stable.yml:3750-3782 | severity=low | confidence=3
  flagged_by: [deepseek_deepseek-v4-pro, moonshotai_kimi-k2_5, qwen_qwen3_6-plus]
  PROBLEM: The prerequisite guard checks that the scripts exist, but not that the consolidator can successfully import its sibling helper.
  WHY: Some reviewers considered the later `||` fallback sufficient, but others pointed out the guard itself does not verify import compatibility.
- scripts/consolidate_soft_error_reports.py:93-140 | severity=low | confidence=5
  PROBLEM: `child.name` and the regex match are left outside the OSError-protected block after `child.is_dir()`.
  WHY: Reviewers disagreed on whether that code can actually raise, but they all focused on the same missing guard around the post-`is_dir()` path.

- .github/workflows/test-and-mark-stable.yml line ~3752
  PROBLEM: SOFT_ERROR_TELEGRAM_BYTES rejects `-1`.
  WHY: The shell regex `^[0-9]+$` only accepts digits, while the reviewer notes the repo’s opt-out convention is `0`.
- .github/workflows/test-and-mark-stable.yml:3750-3782
  PROBLEM: The prerequisite guard only checks file existence, not importability.
  WHY: The reviewer says the shell fallback already covers unexpected failures, so the concern is addressed.
- scripts/consolidate_soft_error_reports.py:93-122
  PROBLEM: `child.name` is unguarded after the OSError handling around `child.is_dir()`.
  WHY: The reviewer argues the path object may still fail in a race.
- tests/test_consolidate_soft_error_reports.py:109-152
  PROBLEM: The new test does not fully validate layout-2 fallback.
  WHY: The reviewer says it claims to cover layout-2 recovery, but the separate `glob` path already handles the OSError case.

- tests/test_consolidate_soft_error_reports.py ~112-132 | severity=medium | confidence=5
  PROBLEM: The monkeypatch-based `test_discover_oserror_on_iterdir_falls_through_to_layout2` test breaks layout-2 fallback.
  WHY: Patching `Path.iterdir` on the root directory also affects the layout-2 path the test expects to recover through.
- .github/workflows/test-and-mark-stable.yml ~3815-3821 | severity=low | confidence=2
  PROBLEM: `SOFT_ERROR_TELEGRAM_BYTES` validation mismatches the Python consolidator’s disable-truncation behavior.
  WHY: The shell regex rejects `-1`, so the workflow falls back to `2800` instead of honoring the sentinel the reviewer expected.
- .github/workflows/test-and-mark-stable.yml:3765-3767
  PROBLEM: The prerequisite guard issue is already addressed.
  WHY: The reviewer notes both scripts are checked now and the Python call has a fallback.
- scripts/consolidate_soft_error_reports.py:126-138 | severity=low | confidence=1
  PROBLEM: The race concern after `iterdir()` is overstated.
  WHY: `child.name` is just path-string access and does not touch the filesystem.

- .github/workflows/test-and-mark-stable.yml:3817 | confidence=3
  PROBLEM: The `SOFT_ERROR_TELEGRAM_BYTES` sanitizer rejects the documented opt-out value.
  WHY: The shell regex only accepts non-negative integers, while the reviewer says the Python consolidator treats `max_bytes <= 0` as disable-truncation.
- .github/workflows/test-and-mark-stable.yml:3765-3767 | confidence=3
  PROBLEM: The prerequisite guard checks file existence but not import compatibility.
  WHY: The reviewer says a future import-path change or syntax error could still break the consolidator before stub outputs are written.
- tests/test_consolidate_soft_error_reports.py:111-132 | confidence=4
  PROBLEM: The monkeypatch in `test_discover_oserror_on_iterdir_falls_through_to_layout2` may inadvertently suppress layout-2 discovery.
  WHY: The reviewer thinks patching `Path.iterdir` can also interfere with the flat-layout recovery path the test wants to exercise.
- scripts/consolidate_soft_error_reports.py:127-140 | confidence=2
  PROBLEM: `discover_reports` does not guard `child.name` and the regex match with the existing OSError handling.
  WHY: The reviewer says the fail-open pattern is inconsistent because the name extraction sits outside the filesystem-error guard.

- tests/test_consolidate_soft_error_reports.py, lines 112-133 | confidence=4
  PROBLEM: `test_discover_oserror_on_iterdir_falls_through_to_layout2` does not exercise what its docstring claims.
  WHY: The reviewer says the layout-2 `glob()` path is not actually getting the same filesystem error simulation.
- scripts/consolidate_soft_error_reports.py, lines 132-134 | confidence=5
  PROBLEM: `child.name` and `ARTIFACT_NAME_RE.match()` cannot raise OSError here.
  WHY: The reviewer says those operations are in-memory path/regex work, so the pass-1 concern is incorrect.
- .github/workflows/test-and-mark-stable.yml, line 3818 | confidence=2
  PROBLEM: The `SOFT_ERROR_TELEGRAM_BYTES` `-1` rejection is a pre-existing inconsistency.
  WHY: The reviewer says the workflow now documents `0` as the disable value, so the mismatch is informational rather than a runtime bug.

- .github/workflows/test-and-mark-stable.yml:3771-3813 | confidence=3
  PROBLEM: The fallback stub hardcodes a crash diagnostic.
  WHY: The reviewer says the `||` branch can run for any non-zero Python exit, so the message can misdescribe the failure mode.
- scripts/consolidate_soft_error_reports.py:109-122 | confidence=2
  PROBLEM: `child.name` is accessed unguarded after the `child.is_dir()` check.
  WHY: The reviewer says the OSError handling covers `is_dir()` and `glob()`, but not the name extraction.
- tests/test_consolidate_soft_error_reports.py:118-140 | confidence=5
  PROBLEM: The `Path.iterdir` monkeypatch breaks the intended layout-2 fallback coverage.
  WHY: The reviewer says the monkeypatch affects the same root path the test relies on for layout-2 recovery.
- tests/test_consolidate_soft_error_reports.py:141-152 | confidence=3
  PROBLEM: The second test does not prove per-child OSError handling.
  WHY: The reviewer says patching `input_dir.is_dir()` does not validate the inner `child.is_dir()` branch the code is meant to harden.

- tests/test_consolidate_soft_error_reports.py:122-145 | confidence=5
  PROBLEM: The `Path.iterdir` monkeypatch breaks the layout-2 fallback test.
  WHY: The reviewer says both layout-1 and layout-2 can hit the same patched path, so the test would not return `clarify` as asserted.
- .github/workflows/test-and-mark-stable.yml:3800-3814 | confidence=3
  PROBLEM: The `SOFT_ERROR_TELEGRAM_BYTES` sanitizer rejects `-1`.
  WHY: The reviewer says the shell gate blocks the Python code’s `max_bytes <= 0` disable-truncation path.
- scripts/consolidate_soft_error_reports.py:132-134 | confidence=5
  PROBLEM: The pass-1 `child.name` race concern is incorrect.
  WHY: The reviewer says `Path.name` and the regex match are in-memory operations and do not touch the filesystem.
- .github/workflows/test-and-mark-stable.yml:3765-3767 | confidence=3
  PROBLEM: The prerequisite-guard concern is already addressed.
  WHY: The reviewer says both scripts are checked and the workflow’s fallback covers unexpected Python failures.

shubhodeep1:
## Claude-branch review

* Commit: `803938d544b191d1b2542e24be1ed8c7c7d666eb`
* Ref: `claude/test-error-reporting-cLOIx`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25167757683

=== CONSENSUS FINDINGS ===
- scripts/truncate_to_utf8_byte_cap.py:60-75 | severity=low | confidence=[3]
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: The helper treats `cap == 0` and `cap < 0` identically by returning the input unchanged, which does not match the CLI’s explicit rejection of negative caps.
  WHY: This creates an inconsistent contract between direct library callers and the CLI entry point; the reviewer notes mixed users could be surprised even though the consolidator’s direct calls avoid negatives.
- .github/workflows/test-and-mark-stable.yml:3988-3993 | severity=medium | confidence=[4]
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: The `wc -c | tr -d '[:space:]'` pipeline is still fragile if `wc` ever emits unexpected whitespace-padded output.
  WHY: Under `set -euo pipefail`, a non-numeric `MSG_BYTES` would make the comparison fail, aborting the Telegram step and potentially leaving the message untruncated.

(No findings reported.)

(No findings reported.)

(No findings reported.)

(No findings reported.)

- scripts/truncate_to_utf8_byte_cap.py:60-75 | severity=low | confidence=[3]
  PROBLEM: Docstring now correctly qualifies the cap<=0 no-op (output may exceed cap), but the implementation treats cap==0 and cap<0 identically without distinguishing the documented CLI rejection of negative caps (exit 2).
  WHY: Library callers passing cap<0 get full input (matching cap==0), but CLI explicitly rejects negatives — inconsistent contracts could surprise mixed users (e.g. consolidator calls library directly).
- .github/workflows/test-and-mark-stable.yml:3988-3993 | severity=medium | confidence=[4]
  PROBLEM: tr -d '[:space:]' strips ALL whitespace including trailing newline from wc -c, but wc -c always emits trailing newline — now MSG_BYTES lacks it, but printf|wc -c|tr chain is still vulnerable if future wc impls embed other whitespace/padding.
  WHY: If wc -c emits unexpected output (e.g. runner image change), MSG_BYTES becomes non-numeric; [ "${MSG_BYTES}" -gt 4000 ] fails under set -euo pipefail, skipping truncation and potentially exceeding Telegram 4096-char cap (API rejection).

(No findings reported.)

shubhodeep1:
## Claude-branch review

* Commit: `6f5ba9fd1bc08398a3c734b0ba61f0066283647e`
* Ref: `claude/test-error-reporting-cLOIx`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25168996760

=== CONSENSUS FINDINGS ===
- scripts/consolidate_soft_error_reports.py:88 | severity=medium | confidence=4
  PROBLEM: `NEXT_HEADING_RE = re.compile(r"^##\s+\S", re.MULTILINE)` can prematurely stop extraction when findings content contains a legitimate `##` heading.
  WHY: At least one reviewer confirmed the parser can truncate later content on real `##` lines, while other reviewers disputed only the `###`-matching rationale.
- scripts/consolidate_soft_error_reports.py:210-211 | severity=low | confidence=4
  flagged_by: [deepseek_deepseek-v4-pro, minimax_minimax-m2_5, moonshotai_kimi-k2_5, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The placeholder check can drop genuine short findings that merely contain phrases like "no findings" or "no issues".
  WHY: Multiple reviewers showed the `len(body) < 160` substring test can incorrectly filter real findings when placeholder phrases appear in context.

- scripts/consolidate_soft_error_reports.py:210 | severity=low | confidence=3
  PROBLEM: The placeholder check can incorrectly filter genuine short findings that contain placeholder phrases as normal prose.
  WHY: The substring-based test against `("no soft errors", "no findings", "none observed", "no issues", "(none)")` will return `""` for a real finding under 160 characters if it contains one of those phrases.

- scripts/consolidate_soft_error_reports.py:88 | confidence=4
  PROBLEM: `NEXT_HEADING_RE` is too broad and can treat legitimate `##`-style issue text inside findings as a new section heading.
  WHY: `extract_findings_section` stops at the first regex hit in the tail, so a line like `## Rate Limit Issue` can truncate the rest of the findings body.
- scripts/consolidate_soft_error_reports.py:~102 | confidence=3
  PROBLEM: Placeholder detection can incorrectly filter legitimate findings that contain placeholder phrases as substrings.
  WHY: The `any(sig in lower and len(body) < 160 ...)` check drops real short findings such as ones containing "no findings" or "none observed" in context.

- scripts/consolidate_soft_error_reports.py:210-211 | severity=low | confidence=4
  PROBLEM: The placeholder detection logic can incorrectly filter genuine short findings that mention placeholder phrases in context.
  WHY: A concise real finding like "No soft errors were found in the rate limiting module" would match `no soft errors` and be removed because the body is under 160 characters.

(No findings reported.)

(No findings reported.)

- scripts/consolidate_soft_error_reports.py:207-211 | confidence=3
  PROBLEM: Placeholder detection can filter out genuine short findings that merely mention a placeholder phrase.
  WHY: The `sig in lower and len(body) < 160` condition can incorrectly drop a real finding like "No soft errors were found in X" even though it is not a placeholder report.

shubhodeep1:
## Claude-branch review

* Commit: `da4bdb5d2d7856644ef616a8460cc930b27fbc93`
* Ref: `claude/test-error-reporting-cLOIx`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25170792689

=== CONSENSUS FINDINGS ===
- scripts/consolidate_soft_error_reports.py:71-93 | confidence=5
  flagged_by: [deepseek_deepseek-v4-pro, moonshotai_kimi-k2_5, qwen_qwen3_6-plus, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: NEXT_HEADING_RE is anchored to the analyzer's literal section headings, coupling the parser to prompt-specific stop markers.
  WHY: Reviewers agree this is an intentional trade-off to prevent truncation from generic `##` matching; the disagreement is only about whether the coupling risk is acceptable.
- scripts/consolidate_soft_error_reports.py:197-236 | confidence=5
  flagged_by: [deepseek_deepseek-v4-pro, moonshotai_kimi-k2_5, qwen_qwen3_6-plus, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: Filtering whitespace-only lines before checking PLACEHOLDER_LINE_RE was alleged to misclassify bodies.
  WHY: Reviewers agree the whole-line anchored placeholder check is correct and does not filter real prose that merely mentions the placeholder phrase.
- tests/test_consolidate_soft_error_reports.py:279-346 | confidence=5
  flagged_by: [deepseek_deepseek-v4-pro, moonshotai_kimi-k2_5, qwen_qwen3_6-plus, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The test `test_pure_placeholder_body_still_filtered_with_bullet` was alleged to lack an assertion on the expected empty output.
  WHY: Reviewers agree the assertion is present and the critique is factually incorrect.

- scripts/consolidate_soft_error_reports.py:74-82 | confidence=3
  PROBLEM: NEXT_HEADING_RE now matches literal section names (Patterns to watch|Likely benign) from the analyzer prompt, creating coupling to the prompt format.
  WHY: VALID OBSERVATION, NOT A BUG; the code intentionally trades this coupling for avoiding the proven truncation bug from generic `##` matching.
- scripts/consolidate_soft_error_reports.py:231-235 | confidence=1
  PROBLEM: Filtering whitespace-only lines before checking whether all lines match PLACEHOLDER_LINE_RE could misclassify bodies.
  WHY: NOT VALID; whitespace-only lines carry no semantic signal and the anchored regex correctly requires every content line to be a pure placeholder.
- tests/test_consolidate_soft_error_reports.py:338-346 | confidence=5
  PROBLEM: The test lacks an assertion on the expected empty output.
  WHY: FACTUALLY INCORRECT; `assert csr.extract_findings_section(md) == ""` is present.

(No findings reported.)

- scripts/consolidate_soft_error_reports.py:74-79 | confidence=3
  PROBLEM: NEXT_HEADING_RE is intentionally anchored to specific section names emitted by `analyze_soft_errors.py`, which is coupling to the analyzer prompt format.
  WHY: NOT A BUG; the code comment documents the design intent and the target headings are treated as a stable contract.
- tests/test_consolidate_soft_error_reports.py:330-337 | confidence=3
  PROBLEM: The pass 1 finding says the test lacks an assertion on the expected empty output.
  WHY: NOT A BUG; the test does contain `assert csr.extract_findings_section(md) == ""`.
- scripts/consolidate_soft_error_reports.py:233-236 | confidence=2
  PROBLEM: The whitespace-only filtering before placeholder matching could misclassify bodies.
  WHY: NOT A BUG; the whole-line placeholder regex correctly requires every non-blank line to be a pure placeholder.

- scripts/consolidate_soft_error_reports.py:71-93 | confidence=5
  PROBLEM: `NEXT_HEADING_RE` was changed from `^##\s+\S` to `^##\s+(?:Patterns to watch|Likely benign)\b`, which couples the parser to the analyzer's downstream headings.
  WHY: Correct fix; the narrower regex prevents premature truncation when findings contain `##` in their body.
- scripts/consolidate_soft_error_reports.py:197-222 | confidence=5
  PROBLEM: The new whole-line placeholder detection could drop legitimate findings that mention placeholder phrases.
  WHY: Correct fix; the anchored regex only filters lines that are entirely placeholder text, not prose mentioning the phrase.
- tests/test_consolidate_soft_error_reports.py:279-346 | confidence=5
  PROBLEM: The new regression test for the bulleted placeholder case was said to be missing an assert.
  WHY: Incorrect; the test includes `assert csr.extract_findings_section(md) == ""`.

- scripts/consolidate_soft_error_reports.py:71-93 | confidence=5
  PROBLEM: Anchoring `NEXT_HEADING_RE` to the two known section headings couples the consolidator to analyzer prompt output.
  WHY: Already satisfied; this is the targeted fix that prevents generic `##` matching from truncating findings containing `##`.
- scripts/consolidate_soft_error_reports.py:197-222 | confidence=5
  PROBLEM: Filtering whitespace-only lines before applying `PLACEHOLDER_LINE_RE` could overfilter real content.
  WHY: Already satisfied; the whole-line regex correctly filters only pure placeholders and keeps findings that merely mention placeholder text.
- tests/test_consolidate_soft_error_reports.py:279-346 | confidence=5
  PROBLEM: The bulleted placeholder regression test was said to be missing an assertion.
  WHY: Already satisfied; the test explicitly asserts `csr.extract_findings_section(md) == ""`.
- .github/workflows/test-and-mark-stable.yml | confidence=5
  PROBLEM: No changes in last run.
  WHY: Status: no issues; the workflow file was unchanged in this run.

- scripts/consolidate_soft_error_reports.py:71-93 | confidence=5
  PROBLEM: `NEXT_HEADING_RE` is anchored to the literal headings emitted by `analyze_soft_errors.py`, creating prompt-format coupling.
  WHY: NOT A BUG; this deliberate change fixes the prior truncation bug caused by matching any `##` heading.
- scripts/consolidate_soft_error_reports.py:197-222 | confidence=5
  PROBLEM: The placeholder check could misclassify bodies because whitespace-only lines are discarded first.
  WHY: NOT A BUG; the regex requires every remaining non-blank line to be entirely placeholder text, so real prose survives.
- tests/test_consolidate_soft_error_reports.py:279-346 | confidence=5
  PROBLEM: The bulleted placeholder test was said to be missing the expected-empty-output assertion.
  WHY: FACTUALLY INCORRECT; the test includes the assertion.

PR #2003, opened 2026-05-03
fix(review_autofix): scope LAST_RUN_DIFF to actual ai-autofix commits

shubhodeep1:
## Summary

- Release v1.1.0 / e2e-smoke-test failed at Phase 4b (`Editor failed to remove bait line E2E_EDITOR_BAIT_25254380200`). Reasoning-effort overrides from PR #1984 / #1993 took effect — the editor ran at `low` per `review_autofix` run 25254574828 — yet still produced empty output across 6 attempts (3 primary + 3 in-step retry).
- Direct review_autofix run 25254574828 evidence (job 74051788318): consolidator emitted a clean `CLASSIFICATION: must-fix` issue with 6 reviewers flagging it (so the prior `analysis/e2e-smoke-failure-25126757724.md` "passthrough/unclassified" hypothesis does NOT apply here). The editor read all the inputs, saw the bait at L4 of the canary via `nl -ba`, then exited cleanly with empty stdout — no `apply_patch`, no Edit, no Write.
- Root cause: `last_run_diff.patch` contained the bait line as a `+1` hunk because `.github/workflows/review_autofix.yml:2263-2271` unconditionally ran `git diff HEAD~1...HEAD`. The bait commit (`[E2E Smoke Test] inject editor bait line ...`) leaked into LAST_RUN_DIFF as if it were a "previous AI autofix run", and the OSCILLATION GUARD block in the editor prompt (`scripts/review_apply_fixes.sh:441-462`) correctly told the model not to re-modify previously-changed hunks without a regression fingerprint + runtime failure path. The model followed the guardrail and bailed.
- Fix: walk back the last 50 commits and use `${SHA}^...${SHA}` for the most recent `[ai-autofix]`/`[judge-fix]` commit (same prefix regex already used by the iteration counter at line 2096-2105 and by `scripts/review_apply_fixes.sh:2838`). If no autofix commit is found, write a "No previous AI autofix run on PR head" placeholder that matches `scripts/review_run_reviewers.sh:249`'s first-iteration sentinel regex.
- Side effect: production runs where someone manually pushed on top of an autofix commit now correctly see the autofix's actual diff (`HEAD~2...HEAD~1`) instead of the manual push's diff. For the common case (HEAD itself is the autofix commit) `${SHA}^...${SHA}` reduces to the existing `HEAD~1...HEAD` semantics — no behavior change for production iteration 2+.

## Test plan

- [x] Local repro: 3 git fixtures (smoke / iteration 2 / manual-push-on-top) all classify correctly with the new walk-back logic.
- [x] `tests/test_review_autofix_last_run_diff_oscillation_guard.py` — 4 contract invariants locked down (walk-back present, prefix regex matches existing convention, `${SHA}^...${SHA}` range used, placeholder matches first-iteration sentinel).
- [x] All other review_autofix contract tests still pass: `test_review_autofix_editor_noop_cascade_contract.py`, `test_review_autofix_phase_transition_contract.py`, `test_review_autofix_reasoning_schedule.py`, `test_review_autofix_merge_precheck.py`, `test_detect_editor_changes_lost.py`.
- [x] YAML loads cleanly via `yaml.safe_load`.
- [ ] e2e-smoke-test gate must pass on next release run (requires a real autofix invocation against a smoke PR — not reproducible locally).

## Notes for reviewers

- Backwards compatibility (CLAUDE.md §6): `LAST_RUN_DIFF_FILE` / `LAST_RUN_CHANGED_FILES_FILE` / `LAST_RUN_DIFF_STAT_FILE` variable names preserved; `FIRST_RUN_NO_PREVIOUS_AI_CHANGES` sentinel preserved; placeholder phrasing kept compatible with the existing `^(No previous AI autofix|Initial run — no previous commit)` regex in `scripts/review_run_reviewers.sh:249`.
- Minimal change set (CLAUDE.md §5): single block in `review_autofix.yml` plus a new contract test. No prompt edits, no script edits, no scope expansion. The OSCILLATION GUARD prompt itself is correct as-is — only its input was wrong.
- Why not include `[ai-merge-resolve]`: the existing `_prior_autofix_count` regex at `scripts/review_apply_fixes.sh:2838` only counts `[ai-autofix]`/`[judge-fix]`. Keeping the same prefix set here for consistency. If we ever want merge-resolve commits to also count toward the OSCILLATION GUARD's "previous AI autofix run", that should be a separate change applied in both places at once.

https://claude.ai/code/session_01NiDi36zsdkHRGkpk8wVRzU

---
_Generated by [Claude Code](https://claude.ai/code/session_01NiDi36zsdkHRGkpk8wVRzU)_

### DIFF
--- a/.github/workflows/review_autofix.yml
+++ b/.github/workflows/review_autofix.yml
@@ -2260,13 +2260,32 @@ jobs:
           fi
           git show --stat HEAD > "${LAST_COMMIT_STAT_FILE}" || true
 
-          if git rev-parse HEAD~1 >/dev/null 2>&1; then
-            git diff "HEAD~1...HEAD" > "${LAST_RUN_DIFF_FILE}" || true
-            git diff --name-only "HEAD~1...HEAD" > "${LAST_RUN_CHANGED_FILES_FILE}" || true
-            git diff --stat "HEAD~1...HEAD" > "${LAST_RUN_DIFF_STAT_FILE}" || true
+          # LAST_RUN_DIFF must reflect actual prior AI autofix work on PR head,
+          # not whatever HEAD~1 happens to be. The downstream OSCILLATION GUARD
+          # in the editor prompt (scripts/review_apply_fixes.sh:441-462) tells
+          # the model to leave any hunk visible in LAST_RUN_DIFF alone unless
+          # it can attach a "regression fingerprint" + "runtime failure path".
+          # The previous unconditional `git diff HEAD~1...HEAD` fed the editor
+          # the diff of impl commits, smoke-test bait commits, and manual
+          # pushes — falsely tripping the guardrail on first-iteration runs.
+          # Smoke run 25254574828 hit this: bait commit was at HEAD,
+          # last_run_diff showed the bait line, OSCILLATION GUARD activated,
+          # editor produced empty output across 6 attempts (3 + 3 retry) at
+          # reasoning=low → Phase 4b bait-removed verification failed.
+          # Walk back recent commits and only use the diff of the most recent
+          # `[ai-autofix]`/`[judge-fix]` commit, matching the existing prefix
+          # convention used by review_apply_fixes.sh:2838 and the iteration
+          # counter at line 2096-2105.
+          LAST_AUTOFIX_SHA="$(git log --format='%H%x09%s' -n 50 2>/dev/null \
+            | awk -F'\t' '$2 ~ /^\[(ai-autofix|judge-fix)\]/ { print $1; exit }' \
+            || true)"
+          if [ -n "${LAST_AUTOFIX_SHA}" ] && git rev-parse "${LAST_AUTOFIX_SHA}^" >/dev/null 2>&1; then
+            git diff "${LAST_AUTOFIX_SHA}^...${LAST_AUTOFIX_SHA}" > "${LAST_RUN_DIFF_FILE}" || true
+            git diff --name-only "${LAST_AUTOFIX_SHA}^...${LAST_AUTOFIX_SHA}" > "${LAST_RUN_CHANGED_FILES_FILE}" || true
+            git diff --stat "${LAST_AUTOFIX_SHA}^...${LAST_AUTOFIX_SHA}" > "${LAST_RUN_DIFF_STAT_FILE}" || true
           else
-            echo "Initial run — no previous commit" > "${LAST_RUN_DIFF_FILE}"
-            echo "Initial run — no previous commit" > "${LAST_RUN_CHANGED_FILES_FILE}"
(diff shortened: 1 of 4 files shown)


shubhodeep1:
Review disposition (commit aa56039 just pushed):

**Copilot — 4 comments, all addressed:**
1. ✅ `git log --first-parent` added to scope walk to PR head's first-parent chain. Local repro confirmed: a smoke-style branch that merged in a `main` containing `[ai-autofix]` previously selected that unrelated SHA from base; with `--first-parent` it correctly returns empty → placeholder. New contract assertion `git log --first-parent in block` locks this down.
2. ✅ Bad citation `scripts/review_apply_fixes.sh:2838` corrected to point at `.github/workflows/review_autofix.yml:2858` (`_prior_autofix_count` block in `Apply fixes with editor model`) — that's where the matching `[ai-autofix]|[judge-fix]` regex actually lives. The script is only 1038 lines.
3. ✅ Same citation fix in the test docstring.
4. ✅ Same citation fix in the test assertion message.

**gemini-code-assist — 3 comments, all skipping with rationale:**

1. **`analysis/workflow-optimization-2026-05-02-9.md` security disclosure**: Out of scope. The file is in this PR's diff because the branch was forked from a point on `main` that's ahead of `stable` by ~10 commits, including `7825734 chore: add workflow log analysis report` which added the file. It is not a change introduced by this PR. The security concern is real and worth raising, but it should be addressed in the PR that originally added the report (or by editing the report on `main`), not here. This PR cannot touch that file without scope creep.

2. **Tabs vs spaces in the new test file**: Skipping. Repo convention for `tests/test_*.py` is tabs, not spaces — `tests/test_review_autofix_editor_noop_cascade_contract.py` is 99 tab-indented lines vs 10 space-indented lines (mixed; the spaces are continuation lines inside parens), and `tests/test_detect_editor_changes_lost.py` is 206 tabs / 0 spaces. The new file follows the established convention. Switching to PEP-8 spaces here would create the only spaces-indented test in the directory and force a wider style migration that is out of scope for an e2e-smoke fix.

3. **`_generate_diff_block` "brittle string slicing"**: Skipping. The function is a near-copy of `_step_block` in `tests/test_review_autofix_editor_noop_cascade_contract.py:54-61` (already merged), which uses the same `text.find("- name: …")` + `text.find("\n      - name:", …)` boundary pattern to extract YAML steps. Reimplementing it with a YAML parser here would either (a) introduce a divergent extraction style or (b) require migrating the existing test too — both out of scope. The "Rule 1: synchronized marker regexes between Apply fixes and Validate" reference is for the EDITOR_NOOP_SUSPICIOUS cascade contract, not this OSCILLATION GUARD contract — different invariants, different file.

CI is now re-running on the new commit; will surface anything that breaks.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2086, opened 2026-05-04
fix(review_autofix): codex log capture + smoke-only editor prompt directive

shubhodeep1:
## Summary
Two related changes targeting the gpt-5.3-codex empty-output failure mode in `review_autofix.yml`'s editor (runs 25305535590 / 25308327160 / 25310399716, all PRs against `tests/e2e_smoke_canary.txt`):

### 1. Codex-log artifact capture on editor empty-output / noop-suspicious
Mirrors the failure-log capture pattern PR #1940 added for `implement.yml` into `review_autofix.yml`'s codex-agent job. Captures per-attempt editor stdout/stderr (`editor_attempt_*.{txt,err}`), the assembled editor / reviewer / conflict-resolver prompts (including the two-pass `reviewer_prompt_pass1.txt`, `reviewer_prompt_pass2.txt`, and `cross_pollination_summary.txt`), the advisory editor inputs (`floor_tags.txt`, `review_issues.txt`, `ledger_status.txt`, `consolidator_raw.txt`, `parser_stats.txt`), the reviewer-fanout outputs (`review_*.{txt,log,err}`, `pass{1,2}_*`, `status_*`, `consensus_pass1.txt`), and `${CODEX_HOME}/log/`. 7-day retention.

Gate is intentionally broader than `failure() || cancelled()`: review_autofix is fail-open on editor empty-output (when codex returns 0-byte stdout the job posts the noop-comment path and exits `success()`), so a plain failure gate would miss exactly the case the artifact exists for. Triggering on `EDITOR_NOOP_SUSPICIOUS=true` or `EDITOR_CHANGES_LOST=true` keeps the capture aligned with the existing `Telegram editor-noop-suspicious` / `editor-changes-lost` alert paths.

Reviewer-stage `~/.codex/log` directories are wiped by the reviewer scripts before this step runs (logged as a follow-up — would require a script-side change in `review_run_reviewers.sh` and `summarize_reviewer_consensus.sh`).

### 2. Smoke-only editor prompt directive (`IS_SMOKE_TEST` gate) — runtime behaviour change
**Heads-up: this changes runtime editor behaviour, not just diagnostics.** Workflow exports `IS_SMOKE_TEST=true` from the `Detect smoke test and tune LLM settings` step's existing `IS_SMOKE` branch; `scripts/review_apply_fixes.sh` conditionally prepends an `=== E2E SMOKE TEST OVERRIDE — READ FIRST ===` block to the editor prompt body when `IS_SMOKE_TEST=true`. The block:
- Names `tests/e2e_smoke_canary.txt` explicitly + the 3-line target shape.
- Mandates "MUST invoke the apply_patch tool" and forbids "Do NOT exit without calling apply_patch" (covers both the silent-empty-completion and the announce-without-invoke variants).
- Restricts edits to the canary file only.
- Reminds the model to emit `Change status: - edited` after the patch.

Production PRs (no `[E2E Smoke Test]` marker) never set `IS_SMOKE_TEST` so the block is empty and the prompt is byte-identical to today's. The static `pre_assembled_static.txt` portion is unchanged so the prompt-cache key is unaffected for production runs.

**Trade-off acknowledged**: the smoke fixture no longer exercises the exact production prompt verbatim. The codex-log artifact above is the parallel detection mechanism for production-side regressions on real PR autofixes.

## Tests
- `tests/test_review_autofix_failure_log_upload.py` — pins the staging-step file list, gate condition, artifact name, retention, and cleanup-step regex guard. Mirrors `tests/test_implement_post_codex_recovery.py::test_failure_log_artifact_upload_contract`.
- `tests/test_review_autofix_smoke_editor_override.py` — pins the `IS_SMOKE_TEST` export, the gate clause, the override section markers, the load-bearing directives ("MUST invoke the apply_patch tool", "Do NOT exit without calling apply_patch", "smoke-test FAILURE"), the named target file, and the override's position before `INPUT FILES`.
- Both wired into `.github/workflows/ci.yml`.

## Test plan
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/review_autofix.yml'))"`
- [x] `bash -n scripts/review_apply_fixes.sh`
- [x] Both new test files pass locally
- [ ] Next E2E smoke: confirm `codex-review-autofix-failure-logs-<run_id>-<attempt>` artifact uploads on EDITOR_NOOP_SUSPICIOUS or EDITOR_CHANGES_LOST and the bait line is removed (smoke override directs the editor to `apply_patch`)
- [ ] On a successful smoke run with the override active, confirm no artifact is uploaded (gate skipped)

## Motivating run
[25310399716](https://github.com/shubhodeep1/coding-workflows/actions/runs/25310399716) — all 6 editor attempts returned 0-byte stdout from `codex exec`; bait line stayed in `tests/e2e_smoke_canary.txt`; e2e gate failed at `verify-bait-removed`. PR #1940's body classifies the same pattern as upstream `gpt-5.3-codex` / OpenRouter unreliability, and the workflow's own `Switch reasoning effort for editor` step documents: *"gpt-5.3-codex … reasons silently, decides not to call apply_patch, and the session ends cleanly with 0 bytes output"*.

https://claude.ai/code/session_0125Gdudj7gc5BcKwFpWhKob


shubhodeep1:
Responding to the four Copilot review comments:

**[L4790 — two-pass reviewer prompts missing]** Verified and fixed in 8eb0853 — `reviewer_prompt_pass1.txt`, `reviewer_prompt_pass2.txt`, and `cross_pollination_summary.txt` are now in the staging glob and pinned in `tests/test_review_autofix_failure_log_upload.py`. The legacy `reviewer_prompt.txt` stays in the list as a fallback for the single-pass path.

**[L1883 — PR title doesn't reflect prompt change]** Updated. Title is now `fix(review_autofix): codex log capture + smoke-only editor prompt directive` and the body has a top-level "Smoke-only editor prompt directive — runtime behaviour change" callout with the trade-off spelled out. Reviewers should see the runtime change is intentional, not snuck in.

**[ci.yml L164 — tests not in mark-stable.yml / test-and-mark-stable.yml]** Not applying. The mark-stable test lists are deliberately minimal (3 tests: `test_orchestrate_lib`, `test_orchestrate_poll_process`, `test_mark_stable_release_tag_refspec_contract`) and the parallel `tests/test_implement_post_codex_recovery.py` — which pins the equivalent contract for `implement.yml` — is also only registered in `ci.yml`. Adding the new tests to mark-stable while leaving the implement-side equivalent there would diverge from the existing convention. If we want every workflow-contract test gated at stable promotion, that's a broader change that should land as its own PR covering all of them at once.

**[review_apply_fixes.sh L186 — smoke override masks the regression]** Fair architectural concern, acknowledged in the updated PR body. The trade-off was explicit in the user's request ("add some custom instructions to the prompt to make the editors job more explicit for the smoke test? to ensure that it doesnt exit without editing the file"). Two mitigations are already in this PR:
1. The codex-log artifact capture (commit 1) fires on real PR autofixes when `EDITOR_NOOP_SUSPICIOUS=true` — those runs do *not* set `IS_SMOKE_TEST` and so see the unmodified production prompt, giving us first-look visibility into production-prompt regressions through actual review_autofix runs rather than only through the smoke fixture.
2. The smoke override is gated on `${IS_SMOKE_TEST:-false}` so the static `pre_assembled_static.txt` portion of the prompt is byte-identical for production — no prompt-cache or behavioural drift outside the smoke path.

Without the override the smoke gate has been failing reliably (runs 25305535590 / 25308327160 / 25310399716, all empty-output) and the upstream model flakiness is unfixable from our side per PR #1940's analysis. Keeping the gate stuck-red for a model behaviour we can't repair from the repo side blocks stable promotion indefinitely; the override unblocks it while the new artifact capture preserves visibility into the production path.

If you'd prefer a "shadow non-smoke smoke" that runs the production prompt on the canary in parallel (without overriding, purely for regression detection), happy to add that as a follow-up — just say the word.

---
_Generated by [Claude Code](https://claude.ai/code)_

shubhodeep1:
## Claude-branch review

* Commit: `8eb08538e6d95e3305b1c1d4dbe384de15f4dbf0`
* Ref: `claude/analyze-job-logs-3fHZE`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25313222796

=== CONSENSUS FINDINGS ===
- .github/workflows/review_autofix.yml:4815 | severity=low | confidence=5
  PROBLEM: The staging step includes a dead `review_*.err` glob that never matches any reviewer-produced files.
  WHY: `scripts/review_run_reviewers.sh` produces `review_*.txt` and `review_*.log` but not `review_*.err`; this is harmless no-op code, with only one reviewer noting it is cosmetic and others noting it is misleading.
- tests/test_review_autofix_failure_log_upload.py:reviewer_prompt_body.txt assertions missing | severity=medium | confidence=5
  flagged_by: [deepseek_deepseek-v4-pro, moonshotai_kimi-k2_5, qwen_qwen3_6-plus, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The contract test does not assert that `reviewer_prompt_body.txt` is staged.
  WHY: The workflow copies `reviewer_prompt_body.txt`, but the test only pins other context files; this leaves a regression gap if staging drops the body file.
- tests/test_review_autofix_failure_log_upload.py:pass1_*.log / pass2_*.log assertions missing | severity=medium | confidence=4
  flagged_by: [deepseek_deepseek-v4-pro, moonshotai_kimi-k2_5, qwen_qwen3_6-plus, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The contract test does not assert `pass1_*.log` or `pass2_*.log`.
  WHY: The workflow stages reviewer per-pass stderr logs, but the test only covers the `.txt` variants, so removing the `.log` globs would go undetected.
- tests/test_review_autofix_failure_log_upload.py:cache_probe_*.txt / cache_probe_*.log assertions missing | severity=medium | confidence=4
  flagged_by: [deepseek_deepseek-v4-pro, moonshotai_kimi-k2_5, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The contract test does not assert `cache_probe_*.txt` or `cache_probe_*.log`.
  WHY: The workflow stages cache-probe diagnostics, but the test does not pin them, leaving a regression gap if those artifacts stop being staged.
- .github/workflows/review_autofix.yml:4823-4824 | severity=high | confidence=4
  PROBLEM: The workflow stages `cache_probe_*.txt` and `cache_probe_*.log` artifacts without redaction.
  WHY: These artifacts were flagged as potentially exposing repository content snippets from semantic-cache probes; one reviewer disputed that they contain repo data, but the highest severity claimed is high and the disagreement is noted.
- .github/workflows/review_autofix.yml:4813-4824 | severity=medium | confidence=3
  flagged_by: [minimax_minimax-m2_5, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The staging step has no explicit size guard before artifact upload.
  WHY: Reviewers noted large reviewer outputs or logs could hit artifact limits and drop diagnostics, though one reviewer rejected this as a bug and treated it as operational rather than correctness-related.
- scripts/review_apply_fixes.sh:179 | severity=medium | confidence=3
  PROBLEM: The smoke override hardcodes the `Changes made:` bullet text.
  WHY: Reviewers warned that if the model phrases the summary differently, downstream verification may mis-detect `EDITOR_CHANGES_LOST`; another reviewer rejected this as intentional contract behavior.
- scripts/review_apply_fixes.sh:147-194 | severity=low | confidence=3
  PROBLEM: The smoke override relies on an unquoted outer heredoc with command substitution, making it fragile to future quoting changes.
  WHY: The current shell construction is intentional and works, but reviewers noted a future refactor could break override emission; one reviewer rejected this as not a current bug.

- .github/workflows/review_autofix.yml:4815 | severity=low | confidence=5
  PROBLEM: The staging step iterates over `review_*.err` even though no reviewer script produces `.err` files.
  WHY: `scripts/review_run_reviewers.sh` only produces `review_<model>.log`, `review_<model>.txt`, `pass1_<model>.log`, `pass2_<model>.log`, `status_*.txt`, and `consensus_pass1.txt`; the `.err` glob is dead code and a harmless no-op.
- tests/test_review_autofix_failure_log_upload.py:function `test_review_autofix_failure_log_artifact_upload_contract()` | severity=medium | confidence=4
  PROBLEM: The contract test omits `reviewer_prompt_body.txt` from its assertions.
  WHY: The workflow stages `reviewer_prompt_body.txt`, so a future refactor removing it would not be caught by the test.
- tests/test_review_autofix_failure_log_upload.py | severity=medium | confidence=3
  PROBLEM: The contract test omits `pass1_*.log`, `pass2_*.log`, and `cache_probe_*` assertions.
  WHY: The workflow stages `pass1_*.log`, `pass2_*.log`, `cache_probe_*.txt`, and `cache_probe_*.log`, but the test only asserts the `.txt` variants, leaving diagnostics unprotected.
- .github/workflows/review_autofix.yml:4882 | severity=low | confidence=1
  PROBLEM: The cleanup regex guard uses `=~` with unescaped variable content.
  WHY: It currently works with GitHub runner paths, but it is fragile if `RUNNER_TEMP` ever contains regex metacharacters; a plain string comparison would be safer.

- .github/workflows/review_autofix.yml:4815 | severity=low | confidence=3
  PROBLEM: The staging step includes a dead `review_*.err` glob that matches no files.
  WHY: `scripts/review_run_reviewers.sh` does not produce `.err` files, so the glob is a harmless no-op.
- tests/test_review_autofix_failure_log_upload.py | severity=medium | confidence=2
  PROBLEM: The contract test does not assert on `pass1_*.log`, `pass2_*.log`, or `cache_probe_*` files.
  WHY: The workflow stages those artifacts, so a future refactor could remove them without breaking the test.
- tests/test_review_autofix_failure_log_upload.py | severity=medium | confidence=3
  PROBLEM: The contract test does not assert on `reviewer_prompt_body.txt`.
  WHY: The workflow stages `reviewer_prompt_body.txt`, but the test does not pin it, leaving a regression gap.
- .github/workflows/review_autofix.yml:4823-4824 | severity=high | confidence=4
  PROBLEM: The staging step copies `cache_probe_*.txt` and `cache_probe_*.log` artifacts that may contain repository content snippets.
  WHY: Uploading these unredacted could expose potentially sensitive repository data to artifact consumers.
- .github/workflows/review_autofix.yml | severity=medium | confidence=3
  PROBLEM: The staging step has no size check before uploading artifacts.
  WHY: Large reviewer outputs or logs could hit artifact limits and cause diagnostics to be dropped.
- .github/workflows/review_autofix.yml:4813-4824 | severity=low | confidence=2
  PROBLEM: The `review_*.err` glob is dead code.
  WHY: `scripts/review_run_reviewers.sh` never produces `.err` files, so the entry only adds a pointless loop iteration and confusion.

- .github/workflows/review_autofix.yml:4815 | severity=low | confidence=4
  PROBLEM: The staging step includes a dead glob pattern `"${PREVIOUS_REVIEWS_DIR}"/review_*.err` that never matches any files.
  WHY: `scripts/review_run_reviewers.sh` produces `.txt`, `.log`, and `status_*.txt` files, but never `.err` files; the glob is a harmless no-op and may mislead future maintainers.
- tests/test_review_autofix_failure_log_upload.py:191-196 | severity=medium | confidence=3
  PROBLEM: The test does not assert on `pass1_*.log` or `pass2_*.log` files even though the workflow staging step captures them.
  WHY: The workflow includes those `.log` files, but the contract test only asserts `pass1_*.txt` and `pass2_*.txt`, so removing the `.log` staging would go undetected.
- tests/test_review_autofix_failure_log_upload.py | severity=medium | confidence=3
  PROBLEM: The contract test omits `cache_probe_*.txt` and `cache_probe_*.log` from its assertions.
  WHY: The workflow captures cache probe files, but the test never pins them, so a future refactor could drop them without failing CI.
- tests/test_review_autofix_failure_log_upload.py | severity=medium | confidence=3
  PROBLEM: The contract test omits `reviewer_prompt_body.txt` from its staging assertions.
  WHY: The workflow copies `reviewer_prompt_body.txt`, but the test only checks `reviewer_prompt.txt`, leaving a regression gap.
- .github/workflows/review_autofix.yml:4813-4824 | severity=medium | confidence=3
  PROBLEM: The staging step has no size guard before uploading artifacts.
  WHY: Very large reviewer outputs or logs could hit artifact limits and drop diagnostics, though this is framed as a low-risk hardening issue.
- scripts/review_apply_fixes.sh:179 | severity=medium | confidence=3
  PROBLEM: The smoke override hardcodes a specific `Changes made:` bullet format.
  WHY: If the model phrases the patch summary differently, downstream validation may incorrectly trigger `EDITOR_CHANGES_LOST`.
- scripts/review_apply_fixes.sh:149-192 | severity=low | confidence=2
  PROBLEM: The smoke override uses an unquoted outer heredoc with command substitution, which is brittle to future refactoring.
  WHY: Quoting the outer heredoc would stop the substitution from executing and break override emission, though the current implementation is correct.

- .github/workflows/review_autofix.yml:4815 | severity=low | confidence=3
  PROBLEM: The `review_*.err` glob in the staging step matches no files.
  WHY: `scripts/review_run_reviewers.sh` produces reviewer `.txt` and `.log` files but never reviewer `.err` files; the entry is a dead no-op.
- tests/test_review_autofix_failure_log_upload.py:~130-152 | severity=medium | confidence=5
  PROBLEM: The contract test is missing `reviewer_prompt_body.txt` from its assertions.
  WHY: The workflow copies `reviewer_prompt_body.txt`, so dropping it from staging would not be caught by the test.
- tests/test_review_autofix_failure_log_upload.py:~155-165 | severity=medium | confidence=4
  PROBLEM: The reviewer-fanout assertion loop omits `pass1_*.log` and `pass2_*.log`.
  WHY: Those logs are captured by the workflow staging step, but the test only covers the `.txt` variants.
- .github/workflows/review_autofix.yml | severity=low | confidence=5
  PROBLEM: The staging step includes a dead `review_*.err` glob.
  WHY: No reviewer script produces `.err` files, so the glob is a pointless no-op that adds iteration overhead.
- tests/test_review_autofix_failure_log_upload.py | severity=medium | confidence=4
  PROBLEM: The contract test omits assertions for `pass1_*.log`, `pass2_*.log`, `cache_probe_*.txt`, and `cache_probe_*.log`.
  WHY: The workflow stages those artifacts, so removing them later would not fail the test and would lose diagnostics.
- tests/test_review_autofix_failure_log_upload.py | severity=medium | confidence=4
  PROBLEM: The contract test omits `reviewer_prompt.txt` and `reviewer_prompt_body.txt` from explicit assertions.
  WHY: The workflow copies both prompt files, but the test does not pin them, leaving prompt reconstruction unprotected.
- .github/workflows/review_autofix.yml | severity=high | confidence=5
  PROBLEM: `cache_probe_*` files are uploaded to artifacts without redaction.
  WHY: The files were asserted to contain raw repository snippets from semantic-cache probes, so uploading them exposes private repo content.
- .github/workflows/review_autofix.yml | severity=medium | confidence=4
  PROBLEM: There is no pre-upload size bound on staged files and logs.
  WHY: A large PR could produce enough review artifacts to hit GitHub artifact limits and truncate diagnostics.
- scripts/review_apply_fixes.sh | severity=medium | confidence=3
  PROBLEM: The smoke override hardcodes the exact `"Changes made:"` bullet phrasing expected after patch application.
  WHY: If the editor phrases the summary differently, downstream verification may falsely trigger `EDITOR_CHANGES_LOST`.
- scripts/review_apply_fixes.sh | severity=medium | confidence=3
  PROBLEM: The smoke override depends on an unquoted outer heredoc for `$()` execution.
  WHY: A future refactor that quotes the heredoc would stop the substitution from running and would emit the override text literally.

- .github/workflows/review_autofix.yml:~4815 (inside staging step for loop) | severity=low | confidence=4
  PROBLEM: The dead glob `"${PREVIOUS_REVIEWS_DIR}"/review_*.err` matches no files.
  WHY: `scripts/review_run_reviewers.sh` produces `.log` files for reviewer output, not `.err` files; the `.err` extension is only used for editor attempts.
- tests/test_review_autofix_failure_log_upload.py | severity=medium | confidence=3
  PROBLEM: The contract test omits assertions for `pass1_*.log`, `pass2_*.log`, `cache_probe_*.txt`, and `cache_probe_*.log`.
  WHY: The workflow stages those artifacts, but the test only asserts `review_*.log` and the `pass*/*.txt` files, so removing them would go undetected.
- tests/test_review_autofix_failure_log_upload.py | severity=medium | confidence=4
  PROBLEM: The contract test omits assertion for `reviewer_prompt_body.txt`.
  WHY: The workflow copies `reviewer_prompt_body.txt`, but the test does not include it in its assertions, leaving a regression gap.

shubhodeep1:
Addressed the claude-branch-review consensus findings in af1c8d8:

**Applied (high-confidence consensus):**
- Removed the dead `review_*.err` glob from the staging step. Six reviewers flagged it; verified that `scripts/review_run_reviewers.sh` produces `review_*.log` (not `.err`) — the `.err` extension is editor-only (`scripts/review_apply_fixes.sh:964`). Added an anti-assertion in the contract test so it can't sneak back.
- Expanded `tests/test_review_autofix_failure_log_upload.py` to pin the missing globs: `reviewer_prompt_body.txt`, `pass1_*.log`, `pass2_*.log`, `cache_probe_*.txt`, `cache_probe_*.log`.

**Investigated, kept as-is:**
- `cache_probe_*` security concern — verified the contents. `cache_probe_prompt.txt` is the literal string `"Return exactly the word CACHE_PROBE_OK."`; `cache_probe_output.txt` is `pre_assembled_static.txt` + that probe prompt; `cache_probe_call_{1,2}.log` is codex stderr from running the probe. `pre_assembled_static.txt` is the cached system instructions block (CLAUDE.md / agents.md / codex_system_instructions.md content) which is all public/checked-in. No private repo data exposure on this path. The 2-of-7 reviewer split on this finding (one reviewer rejected, several flagged; the consensus body explicitly noted the disagreement) lines up with that read.

**Skipped (false positives or out of scope):**
- "Hardcoded `Changes made:` bullet phrasing" (severity=medium, confidence=3, 2 reviewers, 1 dispute) — the smoke override *suggests* a bullet form; the `EDITOR_CHANGES_LOST` detection at `review_autofix.yml:3001-3186` matches any edit verb regex (`modif|chang|updat|add|remov|delet|...`), so the model is free to rephrase. Disputed correctly by one reviewer as intentional.
- "Unquoted outer heredoc fragility" (severity=low, confidence=2-3, 2 reviewers) — same heredoc already uses `$()` for the `LINKED ISSUE` block at `scripts/review_apply_fixes.sh:174-182`; the smoke override follows the existing pattern. Quoting the heredoc would break far more than just the override.
- "No size guard before artifact upload" (severity=medium, confidence=3, 3 reviewers, 1 reject) — operational rather than correctness. `actions/upload-artifact@v6` handles per-file (2GB) and aggregate limits; reviewer outputs are bounded by the script's own truncation. `retention-days: 7` bounds storage cost.
- `cleanup regex guard with unescaped variable content` (severity=low, confidence=1, 1 reviewer) — the regex inputs (`RUNNER_TEMP`, the staging-dir name) are GitHub-controlled and runner-controlled, not user-controlled. Bash `=~` against `${RUNNER_TEMP:-/tmp}/codex-review-autofix-failure-logs$` is the same shape as the equivalent guard PR #1940 added in implement.yml.

---
_Generated by [Claude Code](https://claude.ai/code)_

shubhodeep1:
Addressed three of four new Copilot comments in 04a9b21:

**[scripts/review_apply_fixes.sh L158 — smoke override fired before bait was injected]** Verified: `internal-review.yml:4` triggers `review_autofix` on `pull_request.opened`, which runs the FIRST review_autofix round the moment the smoke PR is created — before `test-and-mark-stable.yml:904-1009`'s Phase 3c injects the bait. On that first round the canary is the legitimate 3-line file and my previous override forced a guaranteed-false `apply_patch`. Fix: gate the override on `IS_SMOKE_TEST=true AND grep -qE '^# E2E_EDITOR_BAIT_[0-9]+:' tests/e2e_smoke_canary.txt` so the override only renders on the post-Phase-3c rounds. Diagnostic echoes name which branch fired so first-round vs. bait-round is grep-able from the live job log.

**[scripts/review_apply_fixes.sh L150 — leading blank line on production prompts]** Verified: the previous `$(if ...; fi)` inside the heredoc expanded to an empty string but left a trailing newline, leaking a blank line in front of `INPUT FILES` on every production run. That shifts the body's bytes/sha and breaks both the OpenRouter prompt-cache key and the `Editor prompt sha256` drift-detection log line on real PR autofixes. Fix: restructure as a brace group:

```bash
{
  if [ "${EDITOR_BODY_RENDER_SMOKE}" = "true" ]; then
    cat <<'__SMOKE_OVERRIDE__'
    ...
__SMOKE_OVERRIDE__
  fi
  cat <<__EDITOR_PROMPT__
INPUT FILES
...
__EDITOR_PROMPT__
} > "${EDITOR_PROMPT_BODY_FILE}"
```

When the smoke gate is false, only the second `cat` fires and the body is byte-for-byte identical to the pre-#2086 implementation. The new contract test `test_production_prompt_body_byte_identical_when_smoke_unset_or_no_bait` runs the actual assembly block in a sandboxed tempdir across both branches and asserts sha256 equality, so a future structural regression back into `$(if …)-in-heredoc` fails CI immediately.

**[.github/workflows/review_autofix.yml L4794 — missing editor inputs in staging]** Verified: `scripts/review_apply_fixes.sh:521` marks `reviewer_manifest.txt`, `reviewer_bundle.txt`, and `last_run_diff.patch` as `required_file` for the editor; the editor's `INPUT FILES` section also names `last_run_changed_files.txt`, `pr_changed_files.txt`, `last_commit_stat.txt`, and `last_run_diff_stat.txt`. All seven added to the staging glob and pinned in `tests/test_review_autofix_failure_log_upload.py`.

**[.github/workflows/ci.yml L164 — tests not in mark-stable lists]** Same response as last round, repeating because Copilot re-flagged it: the parallel `tests/test_implement_post_codex_recovery.py` (which pins the equivalent failure-log contract for `implement.yml`) is also only registered in `ci.yml`, not in `mark-stable.yml` or `test-and-mark-stable.yml`. The mark-stable test lists are deliberately minimal (3 tests) and serve as a fast pre-promotion gate, not a comprehensive contract suite. Adding the new tests there while leaving the implement-side equivalent in ci.yml-only would diverge from established convention. If you want every workflow-contract test gated at stable promotion, that's a broader architectural change (covering all of them at once) that should land as its own PR. Happy to scope that work if you want to make it the new convention.

---
_Generated by [Claude Code](https://claude.ai/code)_

shubhodeep1:
## Claude-branch review

* Commit: `04a9b211f82b9d958eebc8e463ccf46e26be79ec`
* Ref: `claude/analyze-job-logs-3fHZE`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25315301289

=== CONSENSUS FINDINGS ===
- tests/test_review_autofix_failure_log_upload.py:148-227 | severity=medium | confidence=4
  flagged_by: [deepseek_deepseek-v4-pro, minimax_minimax-m2_5, moonshotai_kimi-k2_5, qwen_qwen3_6-plus, z-ai_glm-5]
  PROBLEM: The contract test does not assert all files staged by the workflow, leaving `memory_context.txt` and `original_pr_diff.patch` unprotected from future staging regressions.
  WHY: The staging step copies these files, but the test coverage omits them, so a refactor could silently drop them without CI catching the loss; reviewers disagree on `reviewer_prompt.txt`, so it is not included in the merged root cause.
- .github/workflows/ci.yml:153-168 | severity=low | confidence=5
  flagged_by: [minimax_minimax-m2_5, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The new contract tests are wired only into `ci.yml` and not the stable-branch promotion workflows.
  WHY: This means regressions in these tests could be missed by release-promotion gates; severity is low because reviewers note this follows existing convention and appears to be a workflow-policy gap rather than a runtime defect.
- .github/workflows/review_autofix.yml:4878-4892 | severity=low | confidence=2
  flagged_by: [minimax_minimax-m2_5, x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The cleanup guard uses a regex match with an unquoted RHS variable expansion.
  WHY: The pattern could behave unexpectedly if `${RUNNER_TEMP}` contained regex metacharacters, but reviewers disagree on practical risk and note GitHub-controlled paths make this unlikely.
- .github/workflows/review_autofix.yml:4790-4830 | severity=medium | confidence=3
  flagged_by: [x-ai_grok-4_1-fast, z-ai_glm-5]
  PROBLEM: The staging loop copies `cache_probe_*` artifacts without any size check.
  WHY: Reviewers warn this could allow oversized artifacts, though they also note the files are typically small and bounded in practice.
- scripts/review_apply_fixes.sh:170-174 | severity=medium | confidence=5
  flagged_by: [deepseek_deepseek-v4-pro, z-ai_glm-5]
  PROBLEM: None; this is a rejected finding because the file-existence guard is already present before `grep`.
  WHY: The guard short-circuits the `grep` when the canary file is absent, so the claimed silent-skip bug is not present.
- tests/test_review_autofix_smoke_editor_override.py:291-292 | severity=medium | confidence=5
  flagged_by: [deepseek_deepseek-v4-pro, z-ai_glm-5]
  PROBLEM: None; this is a rejected finding because the empty-file fixtures are intentional for the byte-identical structural test.
  WHY: The test is checking control-flow invariants, and the reviewers explicitly state empty inputs are the correct fixture.

- tests/test_review_autofix_failure_log_upload.py:195-227 | severity=medium | confidence=4
  PROBLEM: The contract test does not include `memory_context.txt` and `original_pr_diff.patch` in its assertion loop, so those staged files are not protected by the test.
  WHY: The workflow stages both files, but the corresponding test loop omits them; a future refactor that drops either file would still pass CI, losing diagnostic context without notice.
- scripts/review_apply_fixes.sh:170-174 | severity=high | confidence=5
  PROBLEM: None; this is a false positive because the canary-file existence check is already present before `grep`.
  WHY: The `[ -f "${CANARY_PATH}" ]` guard short-circuits the `grep` when the file is absent, so the claimed silent skip cannot occur.
- .github/workflows/review_autofix.yml:4820-4821 | severity=high | confidence=5
  PROBLEM: None; this is a false positive because the `cache_probe_*` artifacts are fixed-size and do not scale with PR size.
  WHY: The probe prompt is a fixed literal and the outputs are bounded by the one-shot telemetry probe, so the claimed truncation risk does not exist.
- tests/test_review_autofix_smoke_editor_override.py:291-292 | severity=high | confidence=5
  PROBLEM: None; this is a false positive because empty-file stubs are the correct fixture for the structural invariant under test.
  WHY: The test validates byte-identical prompt assembly when the smoke override does not fire, and the alleged heredoc byte-leak would manifest regardless of file contents.

- tests/test_review_autofix_failure_log_upload.py:148-185 | severity=medium | confidence=4
  PROBLEM: The contract test does not include `reviewer_prompt.txt` and `memory_context.txt` in its file assertions, so a future staging regression would not be caught.
  WHY: The workflow stages those files, but the test does not assert them; if either is removed later, CI would still pass and diagnostics would be incomplete.
- .github/workflows/ci.yml:156-167 | severity=low | confidence=5
  PROBLEM: The new contract tests are added only to `ci.yml` and are absent from the stable-branch promotion workflows.
  WHY: That leaves release promotion without coverage for these tests, though the reviewer notes this matches an existing convention.
- .github/workflows/review_autofix.yml:~4879 | severity=low | confidence=2
  PROBLEM: The cleanup step uses `[[ ... =~ ... ]]` with an unquoted variable inside the regex pattern.
  WHY: If `RUNNER_TEMP` ever contained regex metacharacters, the match could behave unexpectedly, though this is unlikely in practice.

- tests/test_review_autofix_failure_log_upload.py:~148 | severity=medium | confidence=4
  PROBLEM: The contract test omits `memory_context.txt` from the assertion loop even though the workflow stages it.
  WHY: A later refactor could remove that staged file without failing CI, reducing diagnostic coverage for the empty-output failure mode.
- tests/test_review_autofix_failure_log_upload.py:~148 | severity=medium | confidence=1
  PROBLEM: The claim that `reviewer_prompt.txt` is missing from the assertion loop is incorrect.
  WHY: The reviewer states it is already covered by the runtime files list, so only `memory_context.txt` is actually missing.

- tests/test_review_autofix_failure_log_upload.py:~85-122 | severity=medium | confidence=4
  PROBLEM: The contract test’s staged-file assertion loop omits `reviewer_prompt.txt`, `memory_context.txt`, and `original_pr_diff.patch`.
  WHY: The workflow copies those files into the staging directory, but the test does not verify their presence, so a future staging regression could silently pass.

- .github/workflows/ci.yml:153-168 | severity=low | confidence=3
  PROBLEM: The new contract tests are added only to `ci.yml` and are missing from the explicit test lists in `mark-stable.yml` and `test-and-mark-stable.yml`.
  WHY: Release-promotion workflows would skip these tests, allowing regressions in review_autofix diagnostics or smoke override behavior to ship undetected.
- .github/workflows/review_autofix.yml:4878-4892 | severity=low | confidence=2
  PROBLEM: The cleanup guard uses `[[ ... =~ ... ]]` with an unquoted RHS variable expansion.
  WHY: Regex metacharacters in the expanded path could alter matching behavior, though the reviewer considers the practical risk low.
- .github/workflows/review_autofix.yml:4878-4892 | severity=low | confidence=1
  PROBLEM: The regex-based cleanup guard is fragile and should be replaced with string equality.
  WHY: A direct string comparison would avoid regex interpretation entirely and be safer.
- .github/workflows/review_autofix.yml:4790-4830 | severity=medium | confidence=2
  PROBLEM: The staging step copies `cache_probe_*.txt` and `cache_probe_*.log` without truncation or size checks.
  WHY: The reviewer argues large probes could produce oversized artifacts, though no hard evidence of a current failure mode is given.

- tests/test_review_autofix_failure_log_upload.py:148-192 | severity=medium | confidence=4
  PROBLEM: The workflow stages `reviewer_prompt.txt`, `memory_context.txt`, and `original_pr_diff.patch`, but the contract test does not assert these files are present.
  WHY: This creates a coverage gap where a later staging regression could silently remove diagnostic artifacts without failing CI.
- .github/workflows/ci.yml:153-168 | severity=low | confidence=5
  PROBLEM: The new contract tests are wired only into `ci.yml`.
  WHY: The stable-branch promotion workflows do not run them directly, so release gating would miss regressions, though the reviewer says this follows existing pattern.
- .github/workflows/review_autofix.yml:4878-4892 | severity=low | confidence=1
  PROBLEM: The cleanup guard uses an unquoted variable inside a regex match.
  WHY: The reviewer considers the risk low because the same pattern exists elsewhere and runner paths are controlled.
- .github/workflows/review_autofix.yml:4790-4830 | severity=medium | confidence=3
  PROBLEM: The `cache_probe_*` files are copied without size checks.
  WHY: The reviewer notes they are usually small diagnostic outputs, but still records the absence of a guard as a medium-priority hardening issue.
- scripts/review_apply_fixes.sh:146-233 | severity=medium | confidence=2
  PROBLEM: None; this cross-pollination claim is rejected because the smoke override is already fail-safe by design.
  WHY: The `-f` check prevents the `grep` from running on missing files, and unreadable files would surface an error rather than skip silently.
- tests/test_review_autofix_smoke_editor_override.py:full file | severity=medium | confidence=2
  PROBLEM: None; this cross-pollination claim is rejected because empty-file stubs are intentional and appropriate for the test scope.
  WHY: The test targets structural byte-identical behavior, not content-dependent behavior.

shubhodeep1:
Addressed the claude-branch-review consensus in e058d94:

**Applied (5-reviewer consensus, confidence=4):** added `memory_context.txt`, `original_pr_diff.patch`, and `reviewer_prompt.txt` to the test's runtime-files assertion loop. All three are in the workflow staging glob today; before this commit only the comment text mentioned them, the actual `assert` was missing.

**Skipped with reasoning:**

- *ci.yml mark-stable test list (low/5)* — same finding as the previous two rounds. Same answer: `tests/test_implement_post_codex_recovery.py` (the implement-side equivalent failure-log contract test) is also only registered in `ci.yml`. Adding the new tests to mark-stable while leaving the implement-side equivalent in ci.yml would diverge from existing convention. Happy to scope a "register every workflow-contract test in mark-stable" PR if that's the new direction, but I don't want to make a unilateral convention change inside this PR.
- *Cleanup regex guard with unquoted RHS (low/2, 1-of-2 reject)* — the regex shape `^${RUNNER_TEMP:-/tmp}/codex-review-autofix-failure-logs$` mirrors PR #1940's equivalent guard in `implement.yml`. `RUNNER_TEMP` is GitHub-Actions-controlled (paths like `/home/runner/work/_temp` on hosted runners); no user-controlled metacharacters can land there. minimax and z-ai both classify the practical risk as low. No change.
- *cache_probe size guard (medium/3, 1 reject)* — deepseek's rejection (high/5) is correct: the probe input is a fixed literal `Return exactly the word CACHE_PROBE_OK`; outputs are bounded by the one-shot telemetry call shape (a few hundred bytes per `cache_probe_call_{1,2}.log`). No scaling with PR size. No change.

**Explicitly rejected by consensus (no action):**
- *scripts/review_apply_fixes.sh L170-174 silent-skip claim* — `[ -f "${CANARY_PATH}" ]` short-circuits the `grep`; rejected by deepseek (high/5) and z-ai (medium/2).
- *test_review_autofix_smoke_editor_override.py empty-file fixtures* — empty stubs are correct for the structural byte-identical invariant; rejected by deepseek (high/5) and z-ai (medium/2).

---
_Generated by [Claude Code](https://claude.ai/code)_

shubhodeep1:
Addressed in c3ce307:

**[L4785 / test L176 — pr_meta.json + pr_all_comments_context.txt]** Verified and fixed. Both files are declared in the workflow's `Initialize runtime workspace` step (`review_autofix.yml:1103`, `:1109`) and named as required editor inputs in `scripts/review_apply_fixes.sh:152` (the `${PR_META_FILE}` line of the editor's `INPUT FILES` section) and `:159` (the `${PR_ALL_COMMENTS_CONTEXT_FILE}` line). Without them a downloaded artifact can't reconstruct the PR intent or discussion context Codex actually saw. Added to the staging glob and pinned in the contract test.

**[ci.yml L164 — mark-stable test list, fourth round]** Same finding, same answer. The implement-side equivalent contract test (`tests/test_implement_post_codex_recovery.py`) is also only registered in `ci.yml`, not in `mark-stable.yml` or `test-and-mark-stable.yml`. Adding the new tests there while leaving the implement-side equivalent in `ci.yml`-only would diverge from established convention — that's a broader policy change covering all workflow-contract tests at once, scope-creep for this PR. If the policy should change, happy to scope a dedicated PR that covers `test_implement_post_codex_recovery.py`, the new ones here, and any other contract tests in the same category — but I'm not making a unilateral convention shift inside #2086.

---
_Generated by [Claude Code](https://claude.ai/code)_

shubhodeep1:
## Claude-branch review

* Commit: `c3ce307f5da99c0efe660055578751bf62257c16`
* Ref: `claude/analyze-job-logs-3fHZE`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25316638883

=== CONSENSUS FINDINGS ===
- .github/workflows/review_autofix.yml:4844 | severity=low | confidence=4
  flagged_by: [deepseek_deepseek-v4-pro, minimax_minimax-m2_5, moonshotai_kimi-k2_5, x-ai_grok-4_1-fast]
  PROBLEM: `cp -rf "${CODEX_HOME:-$HOME/.codex}/log/." "${STAGE_DIR}/codex-home-log/" 2>/dev/null || true` suppresses all stderr from the copy.
  WHY: A disk-full, permission-denied, or partial-copy failure can leave an incomplete codex-home-log artifact with no indication that the copy failed; reviewers disagree only on whether this is a bug versus hardening, so severity is the highest reported and noted as advisory by some reviewers.
- .github/workflows/review_autofix.yml:4887 | severity=low | confidence=4
  flagged_by: [deepseek_deepseek-v4-pro, minimax_minimax-m2_5, moonshotai_kimi-k2_5, x-ai_grok-4_1-fast]
  PROBLEM: `[[ "${CODEX_REVIEW_AUTOFIX_FAILURE_LOG_DIR}" =~ ^${RUNNER_TEMP:-/tmp}/codex-review-autofix-failure-logs$ ]]` interpolates `${RUNNER_TEMP:-/tmp}` unquoted into a regex pattern.
  WHY: If the expanded temp path contains regex metacharacters, the guard’s matching semantics can change and the cleanup check can become less precise; reviewers disagree only on practical risk, so severity is the highest reported and noted as mostly theoretical.

- .github/workflows/review_autofix.yml:4844 | severity=low | confidence=3
  PROBLEM: `cp -rf "${CODEX_HOME:-$HOME/.codex}/log/." "${STAGE_DIR}/codex-home-log/" 2>/dev/null || true` suppresses all stderr from the copy.
  WHY: A disk-full or permission-denied error would be invisible, though the reviewer treats this as a diagnostic hardening issue rather than a correctness bug.
- .github/workflows/review_autofix.yml:4887 | severity=low | confidence=3
  PROBLEM: `[[ "${CODEX_REVIEW_AUTOFIX_FAILURE_LOG_DIR}" =~ ^${RUNNER_TEMP:-/tmp}/codex-review-autofix-failure-logs$ ]]` interpolates `${RUNNER_TEMP:-/tmp}` unquoted into a regex pattern.
  WHY: Regex metacharacters in the expanded path could alter matching semantics, but the reviewer considers the practical risk negligible on GitHub-hosted runners.
- .github/workflows/review_autofix.yml:4850 | severity=low | confidence=1
  PROBLEM: `cp -rf "${CODEX_HOME:-$HOME/.codex}/log/." "${STAGE_DIR}/codex-home-log/" 2>/dev/null || true` suppresses stderr during codex log staging.
  WHY: The reviewer labels this a hardening-only observability gap and suggests emitting a warning instead of silently discarding copy errors.
- .github/workflows/review_autofix.yml:4887 | severity=low | confidence=1
  PROBLEM: The cleanup regex guard depends on `${RUNNER_TEMP:-/tmp}` expansion semantics.
  WHY: The reviewer notes an empty or unusual `RUNNER_TEMP` could affect the regex/path check, but concludes it is not a practical issue because GitHub Actions sets `RUNNER_TEMP` to a non-empty path.

- .github/workflows/review_autofix.yml:~4815 | severity=medium | confidence=5
  PROBLEM: The staging loop includes a `review_*.err` glob that never matches any files produced by the workflow.
  WHY: Reviewers produce `.log` files, not `.err` files, so the glob is dead code and needlessly iterates every run.
- .github/workflows/review_autofix.yml:~4854 | severity=medium | confidence=4
  PROBLEM: `cp -rf "${CODEX_HOME:-$HOME/.codex}/log/." "${STAGE_DIR}/codex-home-log/" 2>/dev/null || true` suppresses all stderr.
  WHY: Disk-full or permission-denied errors are hidden, so the artifact can contain a partial or truncated codex-home-log dump without indication.
- .github/workflows/review_autofix.yml:~4883 | severity=medium | confidence=4
  PROBLEM: The cleanup guard uses an unquoted `${RUNNER_TEMP:-/tmp}` expansion inside a `=~` regex.
  WHY: Regex metacharacters in the expanded path could alter matching semantics and weaken the path-shape guard.
- .github/workflows/review_autofix.yml:~4756-4870 | severity=medium | confidence=3
  PROBLEM: The staging/upload path has no truncation or size cap before artifact upload.
  WHY: A large artifact could approach GitHub’s artifact limit and be dropped or truncated, leaving post-mortem data incomplete.
- tests/test_review_autofix_smoke_editor_override.py:310-335 | severity=low | confidence=2
  PROBLEM: The byte-identical contract test extracts the prompt body assembly block from `review_apply_fixes.sh` and executes it in a subprocess.
  WHY: A future change that mutates the production prompt body after this block could evade the test and break prompt-cache key stability without a CI signal.

- .github/workflows/review_autofix.yml:4887 | severity=medium | confidence=3
  PROBLEM: `[[ "${CODEX_REVIEW_AUTOFIX_FAILURE_LOG_DIR}" =~ ^${RUNNER_TEMP:-/tmp}/codex-review-autofix-failure-logs$ ]]` expands `${RUNNER_TEMP:-/tmp}` unquoted into the regex pattern.
  WHY: A temp path containing regex metacharacters could change the match behavior and cause the cleanup guard to accept an unintended path.
- .github/workflows/review_autofix.yml:4844 | severity=medium | confidence=3
  PROBLEM: `cp -rf "${CODEX_HOME:-$HOME/.codex}/log/." "${STAGE_DIR}/codex-home-log/" 2>/dev/null || true` suppresses all stderr, including copy failures.
  WHY: A partial or failed codex log copy can leave diagnostic artifacts incomplete with no indication that the capture was not fully successful.

- .github/workflows/review_autofix.yml:4852 | severity=low | confidence=3
  PROBLEM: `cp -rf "${CODEX_HOME:-$HOME/.codex}/log/." "${STAGE_DIR}/codex-home-log/" 2>/dev/null || true` redirects stderr to `/dev/null` and always succeeds.
  WHY: If the codex-home-log copy fails, the artifact may contain partial diagnostics with no signal that the copy was incomplete.
- .github/workflows/review_autofix.yml:4887 | severity=low | confidence=3
  PROBLEM: `[[ "${CODEX_REVIEW_AUTOFIX_FAILURE_LOG_DIR}" =~ ^${RUNNER_TEMP:-/tmp}/codex-review-autofix-failure-logs$ ]]` expands `${RUNNER_TEMP:-/tmp}` unquoted inside the regex.
  WHY: Regex metacharacters in the temp path can change match semantics, weakening the cleanup guard on unusual runners.
- tests/test_review_autofix_failure_log_upload.py:148-227 | severity=low | confidence=3
  PROBLEM: The test does not assert that the `PREVIOUS_REVIEWS_DIR` staging loop includes the reviewer-fanout glob block.
  WHY: A future refactor could remove or narrow those globs without the current test catching the regression.
- .github/workflows/review_autofix.yml:4858-4865 | severity=low | confidence=2
  PROBLEM: The upload step sends the entire STAGE_DIR without any truncation or size cap.
  WHY: A large artifact could hit GitHub’s artifact size limit and be truncated or dropped, losing post-mortem data.
- tests/test_review_autofix_smoke_editor_override.py:580 | severity=low | confidence=5
  PROBLEM: The byte-identical production prompt test is said to capture the assembly block before `render_prompt.sh`, but there is no `render_prompt.sh` in this path.
  WHY: The reviewer concludes the test is correct and the cross-pollination concern is a misunderstanding.

- .github/workflows/review_autofix.yml:~4815 | severity=medium | confidence=5
  PROBLEM: The staging loop includes a dead `"${PREVIOUS_REVIEWS_DIR}"/review_*.err` glob that matches zero files.
  WHY: Reviewers produce `review_<model>.txt` and `review_<model>.log` files, never `review_*.err`, so the loop iterates needlessly and can mislead future debugging.
- .github/workflows/review_autofix.yml:~4844 | severity=medium | confidence=4
  PROBLEM: `cp -rf "${CODEX_HOME:-$HOME/.codex}/log/." "${STAGE_DIR}/codex-home-log/" 2>/dev/null || true` redirects stderr to `/dev/null` unconditionally.
  WHY: Copy failures such as disk full, permission denied, or partial copy are hidden, leaving an incomplete codex-home-log artifact with zero indication of failure.
- .github/workflows/review_autofix.yml:~4883 | severity=medium | confidence=3
  PROBLEM: The cleanup guard uses an unquoted `${RUNNER_TEMP:-/tmp}` expansion inside the `=~` regex RHS.
  WHY: If `RUNNER_TEMP` contains regex metacharacters, the match semantics shift and the stale-dir cleanup can fail or match the wrong path.
- tests/test_review_autofix_smoke_editor_override.py:~580 | severity=medium | confidence=3
  PROBLEM: The byte-identical test validates the prompt body by executing the assembly block in a subprocess before any later prompt-rendering mutation.
  WHY: If a future change mutates the body after this block, the test could still pass while production prompt hashes drift.

- .github/workflows/review_autofix.yml:~4815 | severity=medium | confidence=5
  PROBLEM: The pass-1 dead-`review_*.err` glob finding is not a real issue.
  WHY: The workflow no longer includes `review_*.err` in the staging loop, and the test explicitly asserts it must not be present.
- .github/workflows/review_autofix.yml:4852 | severity=medium | confidence=5
  PROBLEM: The pass-1 stderr-suppression finding is not a bug.
  WHY: The `2>/dev/null || true` pattern is intentional best-effort diagnostic capture and matches the existing `implement.yml` behavior.
- .github/workflows/review_autofix.yml:4887 | severity=medium | confidence=5
  PROBLEM: The pass-1 unquoted-`RUNNER_TEMP` regex finding is not a bug.
  WHY: The pattern is intentionally consistent with `implement.yml`, and `RUNNER_TEMP` is controlled by GitHub Actions rather than user input.
- tests/test_review_autofix_smoke_editor_override.py:~580 | severity=medium | confidence=5
  PROBLEM: The pass-1 claim about the byte-identical test is a misunderstanding.
  WHY: There is no `render_prompt.sh` in this path, and the test correctly compares the prompt-assembly block’s SHA-256 hashes across the intended cases.

shubhodeep1:
claude-branch-review consensus disposition (no code changes this round):

**Consensus findings — skip with reasoning:**

- *L4844 `cp -rf … 2>/dev/null || true` stderr suppression* (4 reviewers, low/4) — z-ai explicitly rejects (confidence=5): *"intentional best-effort diagnostic capture and matches the existing implement.yml behavior"*. The exact same pattern lives in `implement.yml` from PR #1940 (`Stage codex logs for upload` step). Hardening this here without touching implement.yml introduces cross-workflow inconsistency. The right scope for stderr-on-cp-failure observability is a follow-up PR that hardens both workflows together; happy to scope that separately.

- *L4887 unquoted-RHS regex guard* (4 reviewers, low/4) — z-ai explicitly rejects (confidence=5): *"intentionally consistent with implement.yml, and `RUNNER_TEMP` is controlled by GitHub Actions rather than user input"*. Third round of the same finding; same answer. `RUNNER_TEMP` on hosted runners is `/home/runner/work/_temp/...` — no metacharacter risk in practice. Same scope-creep argument as above.

**Per-reviewer findings already disposed:**

- minimax *"`review_*.err` dead glob"* — fixed in af1c8d8 four commits back; z-ai's pass-1 row explicitly notes this is no longer present and the test now asserts its absence. minimax was citing stale data.
- minimax/qwen *"no size cap on staging glob"* — previously rejected by deepseek (high/5): cache_probe inputs are a fixed literal `"Return exactly the word CACHE_PROBE_OK"` and outputs are bounded by the one-shot telemetry shape; reviewer-side outputs are bounded by `review_apply_fixes.sh`'s own truncation; editor outputs are bounded by `EDITOR_MAX_WALL=3300s`. No PR-size scaling.
- minimax/x-ai *"byte-identical test could be evaded by mutations after the assembly block"* — qwen and z-ai both reject (confidence=5 each): *"no `render_prompt.sh` in this path"*, *"the test correctly compares the prompt-assembly block's SHA-256 hashes across the intended cases"*. The test extracts and runs the actual production block; any mutation inside that block IS hashed. Mutations OUTSIDE the block (i.e. in `render_prompt.sh` or the static prefix) are out of scope of this test's invariant by design.
- qwen *"test does not assert reviewer-fanout glob block"* — false; `tests/test_review_autofix_failure_log_upload.py:192-217` asserts `review_*.{txt,log}`, `pass{1,2}_*.{txt,log}`, `status_*.txt`, `consensus_pass1.txt`, `cache_probe_*.{txt,log}` plus the anti-assertion that `review_*.err` must NOT appear.

Net: every consensus finding either repeats prior rounds, mirrors a deliberate implement.yml convention this PR is scoped to follow, or is rejected by other reviewers in the same consensus block. Open to a cross-workflow hardening PR if the stderr/regex concerns are worth elevating to "always-on" — just say.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2196, opened 2026-05-06
fix(codex): unblock apply_patch — broaden ANNOUNCE_EDIT_REGEX, wire model_catalog/trust/full-access across every codex config writer

shubhodeep1:
## Summary

Two related fixes for the recurring "codex narrates `apply_patch` but never invokes it" failure mode that's been wasting full retry budgets across implement, review-autofix, judge, and validate runs.

This PR started narrow (regex broaden) and grew once root-cause analysis turned up an upstream codex config gap that affected every workflow — see commit-by-commit summary below.

## Fix 1 — broaden `ANNOUNCE_EDIT_REGEX` in implement.yml's stuck-intent detector

Implement workflow run on issue #201 (fun-token-multi-chain) burned all 5 retry attempts because the stuck-intent detector's regex missed two real-world patterns from `gpt-5.3-codex`:

- `Applying the one-file edit now by removing the redundant ...override` — "one-file" is not in the closed determiner enum, so the regex couldn't reach the noun `edit`.
- `Implementing the one-file cleanup now by removing the redundant ...override` — `implementing` was not in either gerund verb list and `cleanup` was not in the noun list.

**Effect on attempts 2/3/5:** regex didn't match → `attempt_was_empty=false` → `empty_streak` reset to 0 → the 2-attempt bail at `implement.yml:1497` never fired → loop ran the full 5-attempt budget (~3.7 min of retry sleeps + 5 codex invocations) before exiting `Codex implement failed after 5 attempts`. Attempts 1 and 4 happened to match the existing regex (announce-on-stderr; "Applying the plan…ned"), so the streak never reached the threshold of 2.

**Change:** replace the closed determiner enum with a bounded word-token wildcard (up to 6 intervening tokens between gerund verb and noun); extend the verb list with `implementing|removing|deleting|cleaning`, and the noun list with `cleanup|removal|deletion|override|file`. Plus the post-review `cleaning` symmetry fix in the direct-path branch (6/6 reviewers flagged the asymmetry).

Per the existing comment block: false positives don't regress correctness because the regex only fires on already-empty-delta attempts, so broadening strictly increases recall on a path that's already a confirmed no-op.

## Fix 2 — wire codex config defaults across every config writer

Investigation into why the editor/implement/judge runs produce 0-byte stdout traced to a stack of upstream codex regressions that landed between v0.112 and v0.118+:

1. **`openai/codex#11151` + PR `openai/codex#11238` (rust-v0.112+)** removed the offline `model_info` fallback. Codex now resolves `apply_patch_tool_type` via the bundled `models.json` or remote metadata only. For our setup (`provider = openrouter`, slug `openai/gpt-5.3-codex`), neither path reliably registers `apply_patch_tool_type = "freeform"`, so the model has no `apply_patch` tool to call and produces empty stdout after narrating the intended edit.
2. **`openai/codex#14345` / `#14547` (v0.113+)** added a per-directory "trust this directory" prompt that fires even under `--full-auto` and even under `--dangerously-bypass-approvals-and-sandbox`. In exec mode (no TTY) the prompt cannot be answered, and exec runs can stall or skip tool registration depending on the build.
3. **`openai/codex#19020` / `#16643`** report `apply_patch` hangs / sandbox-refresh failures under `workspace-write`, with multiple users confirming `danger-full-access` resolves them.

**Change applied identically to every "Create Codex config" writer in the repo (9 sites):**

- `model_catalog_json = "<absolute>/codex_model_catalog.json"` when the catalog file exists. Our local catalog already declares `apply_patch_tool_type = "freeform"` for every `openai/gpt-5.*` slug we use, so this re-registers the tool independent of remote metadata. Three writers (`implement.yml`, `clarify.yml`, `plan.yml`) were missing this entirely; the rest already had it.
- `[projects."<absolute-workdir>"] trust_level = "trusted"` block, which is the only documented bypass for the v0.113+ trust prompt at the path codex matches (exact path, no globs — `openai/codex#14599`).
- `approval_policy = "never"` and `sandbox_mode = "danger-full-access"` belt-and-braces with the `--full-auto` CLI flag. Per repo policy: GH runners are ephemeral and carry no sensitive state, so full-access is safe and avoids the workspace-write `apply_patch` hang/refresh paths.

**Files touched (9 codex config writers):**
- `.github/workflows/clarify.yml`
- `.github/workflows/implement.yml`
- `.github/workflows/orchestrate.yml`
- `.github/workflows/orchestrate_clarify_respond.yml`
- `.github/workflows/plan.yml`
- `.github/workflows/review_autofix.yml`
- `.github/workflows/workflow-log-analysis.yml` (3 config writers)
- `scripts/orchestrate_poll_process.sh` (4 config writers)
- `scripts/validate_process.sh` (1 config writer; danger-full-access gated on `GITHUB_ACTIONS=true` per Copilot review — local invocations fall back to `workspace-write`/`on-request`, opt-in via `VALIDATE_FORCE_FULL_ACCESS=1`)

**Inheritors (not separate writers, no change needed):** `validate.yml`, `self_heal_validation.sh`, `implement_diagnose_post_codex_failure.sh`, `review_apply_fixes.sh`, `review_rb_judge.sh`, `review_conflict_resolve.sh`, `review_consolidate.sh` all consume one of the above configs (or sed-patch only `model_reasoning_effort`).

**Intentional override preserved:** `summarize_reviewer_consensus.sh` clones `${CODEX_HOME}` and sed-pins `sandbox_mode = "read-only"` for the summariser. The sed pattern overwrites whatever sandbox value the inherited config carries (workspace-write before, danger-full-access now), so the summariser stays read-only.

## Test plan

Regex contract — `tests/test_implement_post_codex_recovery.py::test_announce_edit_regex_contract` (added in this PR):
- [x] All 4 problematic strings from issue #201 match.
- [x] All 8 backward-compat patterns still match (`apply_patch`, `applying the patch`, `applying a minimal change`, `patching \`foo.py\``, `editing src/main.ts`, `I'll apply the patch`, `will apply the implementation`, `Applying the requested overwrite to tests/e2e_smoke_canary.txt`).
- [x] Minimal per-verb/per-noun cases for every term added in this PR (4 wildcard-noun verbs + 5 wildcard-noun nouns + 10 direct-path verbs including `cleaning \`contracts/FunOFT.sol\``).
- [x] Substring-collision negative controls (`applying the profile`, `applying the overridden behavior`).
- [x] 5 benign control strings stay non-matching.

Sandbox & config wiring:
- [x] `python3 -c "yaml.safe_load(...)"` confirms YAML still parses for all 7 workflow files.
- [x] `bash -n` clean for both touched scripts.
- [x] `tests/test_summarize_reviewer_consensus_sandbox_pin.py` (3 tests) still passes — the summariser's read-only sandbox override is sed-based and rewrites whatever sandbox_mode the inherited config carried, so the new `danger-full-access` default doesn't leak into the summariser sandbox.

## Notes for reviewers

- This PR grew from "regex broaden" to "regex broaden + codex config overhaul" because the regex fix alone wouldn't have helped the `model_catalog_json`-missing case — those runs would still announce-without-emit, just at a different layer. Both fixes were needed to actually unblock `apply_patch` invocation. We deliberately landed multiple fixes simultaneously (per repo policy) to short-circuit the recurring failure; once the failure mode stops recurring we can narrow down which leg was the actual blocker.
- A separate Copilot comment flagged the duplication of the codex-config writer across 9 sites. That refactor (extracting to a shared script / composite action) is left out of this PR — see the comment thread for follow-up.

https://claude.ai/code/session_0176zBtpDpDeD71Ne8cxLnGt


shubhodeep1:
## Claude-branch review

* Commit: `3a21f48dd53337d74ef05a074b62fbd570addc99`
* Ref: `claude/analyze-workflow-logs-Wv0ng`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/25467372173

=== CONSENSUS FINDINGS ===
- .github/workflows/implement.yml:1067 | severity=medium | confidence=5
  PROBLEM: `cleaning` is present in the wildcard-noun verb list but missing from the direct-path verb list, so `cleaning \`foo.py\``-style narration does not match the regex.
  WHY: Multiple reviewers verified the asymmetric alternation and all identified the same root cause; severity elevated to the highest reported level.
- tests/test_implement_post_codex_recovery.py:1645-1659 | severity=medium | confidence=3
  PROBLEM: The contract test only exercises long composite strings and lacks minimal cases for the newly added verbs/nouns, weakening regression coverage.
  WHY: All reviewers described the same coverage gap in the must_match list; severity reflects the highest stated level.
- tests/test_implement_post_codex_recovery.py:1577-1594 | severity=low | confidence=2
  PROBLEM: `_extract_announce_edit_regex` assumes the workflow assignment is a single-line quoted string, making the extraction brittle to YAML refactors.
  WHY: Both reviewers raised the same future-proofing concern about single-line extraction; severity and confidence use the highest reported values.
- .github/workflows/implement.yml:1067 | severity=low | confidence=2
  flagged_by: [x-ai_grok-4_1-fast]
  PROBLEM: The bounded `([^[:space:]]+[[:space:]]+){0,6}` repeat may cause avoidable backtracking on very long non-matching input.
  WHY: Single-reviewer performance concern only, and the reviewer explicitly described it as minor.

- .github/workflows/implement.yml:1067 | severity=low | confidence=3
  PROBLEM: `cleaning` is omitted from the direct-path verb list even though it is present in the wildcard-noun list.
  WHY: The reviewer considered this a real but mostly cosmetic asymmetry with near-zero practical impact.
- tests/test_implement_post_codex_recovery.py:1645-1659 | severity=low | confidence=3
  PROBLEM: The test lacks minimal standalone patterns for the newly added verbs and nouns.
  WHY: The reviewer judged this a real gap in coverage but with limited practical risk because the long strings still exercise the new terms indirectly.
- tests/test_implement_post_codex_recovery.py:1582-1592 | severity=low | confidence=1
  PROBLEM: The reviewer claimed the extraction test could falsely pass on a stale cached value if the regex is split across lines.
  WHY: The reviewer explicitly marked this finding incorrect, stating there is no caching and that such a change would fail the test rather than pass silently.
- .github/workflows/implement.yml:1067 | severity=low | confidence=1
  PROBLEM: The reviewer raised a greedy-repeat performance concern for the regex.
  WHY: The reviewer explicitly rejected this as overblown and not a real concern.

- .github/workflows/implement.yml:1067 | severity=high | confidence=4
  PROBLEM: `cleaning` is missing from the direct-path verb list, so `cleaning file.txt`-style narration can fail to match when the wildcard path does not apply.
  WHY: The reviewer tied the issue to the PR intent and treated it as runtime-affecting because the direct path is required for zero-token cases.
- tests/test_implement_post_codex_recovery.py:1645-1659 | severity=medium | confidence=3
  PROBLEM: The contract test lacks minimal coverage for the new verbs and nouns, so regressions can slip through if the long composite strings still match via other subpatterns.
  WHY: The reviewer saw this as a regression-protection gap rather than a direct runtime failure.
- tests/test_implement_post_codex_recovery.py:1577-1585 | severity=low | confidence=2
  PROBLEM: The regex extraction assumes the assignment is on exactly one line with no embedded quotes.
  WHY: The reviewer described this as a valid future concern but not critical.
- .github/workflows/implement.yml:1067 | severity=low | confidence=2
  PROBLEM: The regex’s greedy quantifier could cause quadratic behavior on very long stdout.
  WHY: The reviewer labeled it a minor concern and also noted the bounded quantifier reduces the risk.

- .github/workflows/implement.yml:1067 | severity=high | confidence=5
  PROBLEM: `cleaning` is missing from the direct backtick-path verb list, so `cleaning \`foo.py\`` does not match the regex.
  WHY: The reviewer stated this directly contradicts the PR description and makes the behavior inconsistent with the other verbs.
- tests/test_implement_post_codex_recovery.py:1648-1661 | severity=medium | confidence=3
  PROBLEM: The must_match list only uses long issue-derived strings and does not include minimal new verb/noun cases.
  WHY: The reviewer said this weakens regression protection because a future regression could still match through other alternations.

- .github/workflows/implement.yml:1067 | severity=medium | confidence=4
  PROBLEM: `cleaning` is added to the wildcard-noun verb list but not to the direct-path verb list, so `cleaning \`contracts/FunOFT.sol\`` does not match.
  WHY: The reviewer verified the mismatch and described the resulting retry-budget waste on stuck-intent output.
- tests/test_implement_post_codex_recovery.py:1645-1659 | severity=medium | confidence=3
  PROBLEM: The must_match list contains only long-form strings for the new patterns and no minimal atomic tests for each added verb or noun.
  WHY: The reviewer said a regression in one verb or noun could go undetected if the long strings still match through different alternation paths.
- tests/test_implement_post_codex_recovery.py:1589-1595 | severity=low | confidence=1
  PROBLEM: `_extract_announce_edit_regex` assumes a single-line `var='...'` shell assignment and would fail if the workflow switches to folded YAML.
  WHY: The reviewer labeled this as a hardening concern only and suggested a comment or alternate extraction path.

- .github/workflows/implement.yml:1067 | severity=medium | confidence=4
  PROBLEM: `cleaning` is included in the wildcard-noun gerund list but omitted from the direct-filename gerund list.
  WHY: The reviewer said narration like `cleaning \`contracts/FunOFT.sol\`` can miss the bail trigger and waste retries.
- tests/test_implement_post_codex_recovery.py:1645-1659 | severity=medium | confidence=3
  PROBLEM: The must_match list covers historical and issue-derived strings but omits minimal new-verb/noun combinations.
  WHY: The reviewer said this allows regressions in the broadened recall to slip through.
- tests/test_implement_post_codex_recovery.py:1583-1594 | severity=medium | confidence=4
  PROBLEM: The extraction regex assumes a single-line assignment with no embedded quotes or YAML folding.
  WHY: The reviewer said a future workflow refactor could make extraction fail or become misleading.
- .github/workflows/implement.yml:1058-1060 | severity=low | confidence=2
  PROBLEM: The bounded repeat is described as greedy and potentially slow on long no-op stdout.
  WHY: The reviewer treated this as a minor performance concern.

- .github/workflows/implement.yml:1067 | severity=medium | confidence=4
  PROBLEM: `cleaning` was added to the wildcard-noun verb list but omitted from the direct-path verb list.
  WHY: The reviewer said this creates asymmetric coverage and causes `cleaning \`contracts/FunOFT.sol\`` to miss the direct path.
- tests/test_implement_post_codex_recovery.py:1645-1659 | severity=medium | confidence=3
  PROBLEM: The test covers only specific historical and issue-derived strings, leaving no minimal isolation tests for the newly added verbs and nouns.
  WHY: The reviewer said the current strings can still pass through other subpatterns, weakening regression detection.

PR #2294, opened 2026-05-08
AI implementation for issue #2291

shubhodeep1:
Automated implementation. Closes https://github.com/shubhodeep1/coding-workflows/issues/2291

### DIFF
--- a/.github/workflows/implement.yml
+++ b/.github/workflows/implement.yml
@@ -1056,7 +1056,7 @@ jobs:
               --semble-bin "$(command -v semble 2>/dev/null || true)" \
               --semble-index "${SEMBLE_INDEX_DIR:-}" \
               --semble-query-from "${TARGETED_FILE_CONTEXT_QUERY_FILE:-}" \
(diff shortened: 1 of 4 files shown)


shubhodeep1:
### AI Issue Summary

- Add phase-2 Semble support for implement-phase targeted file context overflow, building on the existing Semble foundation already in place.
- Extend `scripts/targeted_file_context.py` and the implement workflow to pass optional Semble inputs, including a stable query source derived from implement context.
- When oversized targeted-file prompts overflow and Semble is enabled, emit bounded `chunk-retrieved via semble` blocks instead of the legacy overflow-only marker.
- Preserve the current legacy overflow marker as the default fallback whenever Semble is disabled or any helper/query step fails.
- Keep the static prompt prefix unchanged, preserve all non-overflow behavior byte-for-byte, and avoid expanding scope into reviewer, validate, judge, or workflow-template changes.
- Update tests to cover both successful Semble overflow output and disabled/fallback behavior while ensuring existing tests still pass with the feature off.

PR #3072, opened 2026-06-03
fix(stall-recovery): direct in-flight-run fallback before retrigger_review empty-commit push

shubhodeep1:
## Summary

The `retrigger_review` stall recovery skips its empty-commit push when an in-flight review run is detected on the PR head branch — pushing under a still-editing `review_autofix` run trips that run's `AUTOFIX_PRE_EDITOR_STALE_BASE → soft_exit` and discards a full review pass. That in-flight check reads the per-tick `_load_actions_runs_cached` blob, the **same source `build_active_issue_set` consumes**, which the active-set guard comment already documents can miss a live run (cache TTL / 304-reuse, the per-status 50-item listing window, pagination, or `head_branch=null` on `workflow_dispatch`). A false negative there is uniquely destructive at this one push site — it's the failure mode behind a consumer-repo stall incident (hylifegroup.com, PR 12 / issue 11): an orchestrator empty commit landed mid-run and threw away ~50 min of reviewer/editor work.

## What changed

- **New helper `_direct_inflight_review_run_on_branch`** (next to `_load_actions_runs_cached`): on the cache-miss path only, just before the destructive push, it confirms against an authoritative branch-scoped `gh run list --branch` listing (server-side `--branch` filter, so it is **not** subject to the global 50-item listing cap the cached blob is) and reports a fresh `in_progress`/`queued` `AI Review` / `Internal Review` / `Review Autofix` run.
- **Wired into both push sites** — `execute_stall_recovery_action` and `run_standalone_stall_recovery`. (The original report only flagged the first; the second carried the identical gap.)
- If the fallback finds a live run, the push is skipped with `STALL_RECOVERY_EFFECTIVE_ACTION=retrigger_review_skipped_inflight` (logged `direct check — cached scan missed it`).

## Compliance / safety

- **§15 (API hygiene):** fires only on the cache-miss path immediately before the push (the fail-open cache-miss fallback §15 sanctions), so the steady state adds **zero** API calls; a single branch-scoped call, not the 3-call per-workflow loop. Audited and documented why the cached blob and the adjacent failed-autofix loop can't supply the data.
- **Freshness** mirrors the cached scan's `STALL_THRESHOLD_MINUTES` window, so a genuinely hung run doesn't block recovery forever.
- **Fails open** on any `gh`/`jq`/`date` error. **No identifiers renamed (§6).**
- **Known limitation** (documented in the helper + README): `gh run list --json` exposes `workflowName`/`name` but not the workflow file path, so a consumer that renamed the review workflow's display name is matched only if `workflowName` still resolves; on a miss the guard fails open — no worse than pre-fix behaviour.

## Testing

- New `tests/test_retrigger_inflight_direct_fallback.py` (9 cases): detection, freshness gating, non-review/unrelated-branch filtering, queued/`null` `startedAt` coalescing, `workflowName` matching, and fail-open paths. Registered in `ci.yml`, `mark-stable.yml`, `test-and-mark-stable.yml`.
- Regression: the 11 retrigger/standalone tests in `test_orchestrate_poll_process.py` plus the existing stall suites (`test_stall_recovery_pr_lookup`, `test_stall_freshness_branch_fallback`, `test_stall_loop_and_churn_fixes`) all pass. The change is a transparent no-op in the existing integration harness (its `gh run list` stub omits the `name`/`workflowName` fields), so existing flows are unaffected.
- README stall-recovery section documents the new guard.

## Scope notes

This is **Fix B** from an investigation of the stall-recovery incident. Two related items are intentionally **not** in this PR:
- **Fix A** (raise `STALL_THRESHOLD_DONE_MINUTES`, lower `MAX_STALL_RECOVERIES_DONE`, set `AUTOFIX_SKIP_SELF_TRIGGERED=true`) is a set of **consumer-repo repository variables**, not a code change here.
- Whether to promote any of those values to **workflow-wide defaults** is under discussion (2 of the 3 are deliberate existing defaults); a follow-up commit may be added to this branch if we decide to.

https://claude.ai/code/session_01YbKThUNMm3NXBi3YATFZFd

---
_Generated by [Claude Code](https://claude.ai/code/session_01YbKThUNMm3NXBi3YATFZFd)_


shubhodeep1:
## Claude-branch review

* Commit: `2d45821ffe6cf50d483b11ef8007928f25a6fc77`
* Ref: `claude/keen-hypatia-bX9Am`
* Run: https://github.com/shubhodeep1/coding-workflows/actions/runs/26859703127

=== CONSENSUS FINDINGS ===
- tests/test_review_semble_contract.py:102 | confidence=[5]
  flagged_by: [moonshotai_kimi-k2_5]
  PROBLEM: The test expects OPTIONAL_BOOTSTRAP_SCRIPTS to equal "install_semble.sh build_semble_wrapper.sh semble_helpers.sh" but the workflow was updated in commit a904c88 to include "render_prompt.py"
  WHY: The failure is reported at tests/test_review_semble_contract.py line 102, and .github/workflows/review_autofix.yml line 1279 now shows OPTIONAL_BOOTSTRAP_SCRIPTS="install_semble.sh build_semble_wrapper.sh semble_helpers.sh render_prompt.py", so the assertion is stale relative to the current workflow.

=== CONSENSUS TASK GAPS ===
(No task gaps reported.)
=== END CONSENSUS TASK GAPS ===

(No findings reported.)

(No findings reported.)

- tests/test_review_semble_contract.py:102 | confidence=5
  PROBLEM: The test expects OPTIONAL_BOOTSTRAP_SCRIPTS to equal "install_semble.sh build_semble_wrapper.sh semble_helpers.sh" but the workflow was updated in commit a904c88 to include "render_prompt.py"
  WHY: The current PR does not modify review_autofix.yml or OPTIONAL_BOOTSTRAP_SCRIPTS; the mismatch is reported at tests/test_review_semble_contract.py line 102 and the updated value is shown in .github/workflows/review_autofix.yml line 1279.

Issue #3293, opened 2026-06-12
Stage `review-consolidator.txt` into workflow support files


shubhodeep1:
Summary
The source doc's slow-review recommendation mixes two ideas: a safe staging-gap fix and a riskier editor-prompt reshaping change. Current HEAD confirms the safe staging-gap part is still real: `prompts/review-consolidator.txt` exists and `scripts/review_consolidate.sh` already reads `${SUPPORT_PROMPTS_DIR:-prompts}/review-consolidator.txt`, but `scripts/stage_workflow_support.sh` does not copy that prompt into staged support files.

Acceptance criteria
- Teach `scripts/stage_workflow_support.sh` to install `prompts/review-consolidator.txt` from `.codex-workflow-src` or `.codex-workflow-src-main` into `${SUPPORT_PROMPTS_DIR}` using the same optional-support-file pattern already used for `review-reviewer-checklist.txt`.
- If the source prompt is absent, keep fail-open behavior and emit a warning consistent with the existing checklist staging branch.
- Add focused regression coverage for the staging contract.

Out of scope
- Do not redesign `scripts/review_consolidate.sh`.
- Do not change review-model fan-out, editor reasoning, prompt condensation, or parser fail-open semantics here; the closeout report will record those broader ideas separately.

Constraints
- Stay inside `files_touched`.
- Prefer the smallest append-only change in the staging helper; avoid unrelated support-file churn.

---
**Orchestrator metadata** (do not edit)
- Tracking issue: #3291
- Integration branch: orchestrator/project-3291
- Local ID: `review-consolidator-staging-gap`
- Priority: 2
- Managed by: AI Orchestrator

PR #3296, opened 2026-06-12
AI implementation for issue #3293

shubhodeep1:
Automated implementation. Closes https://github.com/shubhodeep1/coding-workflows/issues/3293
Refs #3291

### DIFF
--- a/.ai/.workspace_source_manifest.txt
+++ b/.ai/.workspace_source_manifest.txt
@@ -96,7 +96,7 @@ ai_pipeline.md
 analysis/last_collection_timestamp.txt
 analysis/recommendation-processing-report.md
 analysis/validation-selftest-status.json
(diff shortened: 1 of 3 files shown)


shubhodeep1:
### AI Issue Summary
- Fix a staging gap where `review-consolidator.txt` exists and is already consumed by review consolidation, but is not copied into staged workflow support files.
- Update workflow support staging to include that prompt from the workflow source locations, following the existing optional support-file pattern.
- Preserve fail-open behavior: if the source prompt is missing, emit a warning instead of failing.
- Add focused regression coverage for the staging contract.
- Keep scope tight: stay within `files_touched`, prefer a minimal append-only change, and avoid redesigning review consolidation or broader review-model/editor behavior.

